PR #26, opened 2018-06-19
[fixes #7]: correct access to disconnect event

marcmo:
with this disconnect works again


marcmo:
wrong related issue number

PR #38, opened 2018-07-28
Fix for Future<Null> errors in Dart 2

t04glovern:
Hi,

Running the library with Dart 2 results in a couple compatibility problems. I'm aware that running with the following does work (`flutter run --no-preview-dart-2`), however I fix moving forward could be potentially handled by the changes I've made here.

Happy to discuss though as I'm not certain of the implications of this.

### DIFF
--- a/lib/source/lib_core.dart
+++ b/lib/source/lib_core.dart
@@ -34,18 +34,18 @@ class FlutterBleLib {
     });
   }
 
-  Future<Null> createClient(String restoreStateIdentifier) async {
+  Future<dynamic> createClient(String restoreStateIdentifier) async {
     await _mainMethodChannel.invokeMethod(_createClient,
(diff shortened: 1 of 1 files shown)


ram231:
Would like to fork yours, The project might be not maintained by the author.

t04glovern:
@ram231 We had a lot of success with the lib after using this change in a recent weekend project. Let me know if you run into problems with the fork

duncanfinney:
Fork seems to be working pretty well.

pawelByszewski:
@duncanfinney @t04glovern @ram231 It has been a long time since we stopped maintaining the library but now we come with great news! We are currently working with Frontside on a new BLE Flutter library and BLE simulator. Beta version has been released already and we are going to release 2.0.0 in the first half of November

PR #45, opened 2018-09-02
change int to a double in monitorCharacteristic

leeweisberger:


pawelByszewski:
@leeweisberger It has been a long time since we stopped maintaining the library but now we come with great news! We are currently working with Frontside on a new BLE Flutter library and BLE simulator. Beta version has been released already and we are going to release 2.0.0 in the first half of November

PR #74, opened 2019-09-02
travis Ci config

pawelByszewski:


.travis.yml
```diff
@@ -0,0 +1,7 @@
+git checlanguage: generic
+before_script:
+  - git clone https://github.com/flutter/flutter.git -b stable
+  - export PATH=`pwd`/flutter/bin:`pwd`/flutter/bin/cache/dart-sdk/bin:$PATH
+script:
+  - flutter packages get
+  - flutter test
```
dariuszseweryn: New line? ;)

mikolak: There's no building of the native parts and we should definitely check that on CI.

pawelByszewski: It is not so obvious how to set up a pure native build. I added a ticket for iOS https://github.com/Polidea/FlutterBleLib/issues/75

PR #80, opened 2019-09-10
[android] add MBA as dependency in FlutterBleLib

pawelByszewski:


android/build.gradle
```diff
@@ -25,10 +25,33 @@ android {
...
     defaultConfig {
-        minSdkVersion 16
+        minSdkVersion 18
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
     lintOptions {
         disable 'InvalidPackage'
     }
+
+//    buildTypes {
```
mikolak: Why is it commented?

android/gradle.properties
```diff
@@ -1,4 +1,4 @@
 org.gradle.jvmargs=-Xmx1536M
 
-android.useAndroidX=true
-android.enableJetifier=true
+android.useAndroidX=false
```
mikolak: why?

PR #93, opened 2019-09-16
Scanning feature

bartoszwilk:


android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
```diff
@@ -76,4 +90,30 @@ private void destroyClient(Result result) {
...
+                        try {
+                            scanningStreamHandler.onScanResult(data);
+                        } catch (JSONException e) {
+                            e.printStackTrace();
+                        }
+                    }
+                }, new OnErrorCallback() {
+                    @Override
+                    public void onError(BleError error) {
+                        Log.d("scanning_error", error.toString());
```
dariuszseweryn: Shouldn't the handler informed of error?

lib/scan_result.dart
```diff
@@ -0,0 +1,70 @@
+part of flutter_ble_lib;
+
+abstract class _Metadata {
+  static const String ID = "id";
```
mikolak: https://dart.dev/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names

android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
pawelByszewski: `log.e` ?

android/src/main/java/com/polidea/flutter_ble_lib/converter/ScanResultJsonConverter.java
```diff
@@ -0,0 +1,92 @@
...
+        root.put(Metadata.MTU, value.getMtu());
+        if (value.getAdvertisementData() != null) {
+            serializeAdvertisementData(root, value.getAdvertisementData());
+        }
+        root.put(Metadata.IS_CONNECTABLE, JSONObject.NULL);
+        root.put(Metadata.OVERFLOW_SERVICE_UUIDS, JSONObject.NULL);
+        return root.toString();
+    }
+
+    private void serializeAdvertisementData(JSONObject root,
```
pawelByszewski: what about extracting a few smaller private methods for sake of readability?

android/src/main/java/com/polidea/flutter_ble_lib/converter/ScanResultJsonConverter.java
```diff
@@ -0,0 +1,92 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+public class ScanResultJsonConverter implements JsonConverter<ScanResult> {
+
+    private interface Metadata {
```
pawelByszewski: Do all those data (rssi, mtu, etc.) are connected only with the scan result?

dariuszseweryn: To be honest `mtu` should be associated only with the connection. Not with ScanResult

mikolak: We don't have a `Connection` object. Best we could do is put it in the device, though I suppose it's unknown on iOS and unknown until a request to change it on Android?

lib/lib_core.dart
```diff
@@ -34,4 +37,28 @@ class FlutterBleLib {
     await _methodChannel.invokeMethod(MethodName.DESTROY_CLIENT);
     return;
   }
+
+  Stream<ScanResult> startDeviceScan(int scanMode,
+      int callbackType,
+      List<String> uuids,) async* {
+    await _methodChannel.invokeMethod(
```
mikolak: I think we might lose a few first results, if they come in instantaneously? We're waiting for the future to resolve before we subscribe to the event channel.

android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
bartoszwilk: I will add error handling in another PR. We will send all errors from here straight to the scanningStreamHandler and then we will deserialise it on flutter side.

lib/lib_core.dart
bartoszwilk: Fixed.

lib/scan_result.dart
bartoszwilk: For the sake of conciseness I renamed all const fields' names to match lowerCamelCase.

android/src/main/java/com/polidea/flutter_ble_lib/converter/ScanResultJsonConverter.java
bartoszwilk: Done.

android/src/main/java/com/polidea/flutter_ble_lib/converter/ScanResultJsonConverter.java
bartoszwilk: This contract was copy-pasted from the JS code base. Do we want to change it now?

bartoszwilk: @pawelByszewski AFAIK rssi should be contained only in ScanResult because we can read it's real value only during scanning operation.

bartoszwilk: I submitted an issue to remove MTU from ScanResult class in MBA.

PR #98, opened 2019-09-16
Feature: peripheral connection, connection check, observing connection state, disconnecting

mikolak:


android/src/main/java/com/polidea/flutter_ble_lib/SafeMainThreadResolver.java
```diff
@@ -0,0 +1,44 @@
...
+    private AtomicBoolean called = new AtomicBoolean(false);
+
+    public SafeMainThreadResolver(OnSuccessCallback<T> onSuccessCallback, OnErrorCallback onErrorCallback) {
+        this.onErrorCallback = onErrorCallback;
+        this.onSuccessCallback = onSuccessCallback;
+    }
+
+    public void onSuccess(final T data) {
+        if (onSuccessCallback != null && called.compareAndSet(false, true)) {
+            new Handler(Looper.getMainLooper()).post(new Runnable() {
```
pawelByszewski: This kind of construction may lead to a memory leak. I tried to trace all usages and it seems to be ok.  DO we have anywhere access to activity?

example/lib/main.dart
```diff
@@ -21,41 +25,93 @@ class _MyAppState extends State<MyApp> {
...
+      setState(() {
+        this.connectionState = connectionState;
+      });
+    });
+
+    setState(() {
+      deviceConnectionAttempted = false;
+    });
+
+//    await peripheral.disconnectOrCancelConnection();
```
pawelByszewski: remove commented code

android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
```diff
@@ -64,6 +76,12 @@ public void onMethodCall(MethodCall call, Result result) {
             case MethodName.STOP_DEVICE_SCAN:
                 stopDeviceScan(result);
                 break;
+            case MethodName.CONNECT_TO_DEVICE:
```
pawelByszewski: what about asking a delegate if it is able to handle given method?

android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
```diff
@@ -64,6 +76,12 @@ public void onMethodCall(MethodCall call, Result result) {
...
+            case MethodName.IS_DEVICE_CONNECTED:
+            case MethodName.OBSERVE_CONNECTION_STATE:
+            case MethodName.CANCEL_CONNECTION:
+                deviceConnectionDelegate.onMethodCall(call, result);
```
pawelByszewski: why does the connection have separate delegate but scan does not?

android/src/main/java/com/polidea/flutter_ble_lib/SafeMainThreadResolver.java
mikolak: No, the library uses only application context.

android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
mikolak: We kind of forgot about it. I can move scanning to a separate delegate in next PR.

android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
mikolak: Like a cascade of switches?
`if (xDelegate.canHandle(call))
    xDelegate.handleCall(call, result)
else if (yDelegate.canHandle(....
```

Comment (pawelByszewski) [reply android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java]: `xDelegate.canHandle(call)` exactly.
But I am not sure `if.....else if` is the best solution. There are a lot of possibilities e.g.  to have a list of delegates and iterate over the list

File: android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java (hunk 4 above)
Comment (pawelByszewski) [reply android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java]: 👍

File: android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java [hunk 5]
```diff
@@ -0,0 +1,196 @@
...
+                isDeviceConnected(deviceId, result);
+                return;
+            case MethodName.OBSERVE_CONNECTION_STATE:
+                observeConnectionState(deviceId, call.<Boolean>argument(ArgumentKey.EMIT_CURRENT_VALUE), result);
+                return;
+            case MethodName.CANCEL_CONNECTION:
+                cancelConnection(deviceId, result);
+                return;
+            default:
+                throw new IllegalStateException("");
```
pawelByszewski: `IllegalArgumentException` or some kind `Unsupported....` or some custom exception?

android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java
```diff
@@ -0,0 +1,196 @@
...
+                },
+                new OnErrorCallback() {
+                    @Override
+                    public void onError(BleError error) {
+                        result.error(String.valueOf(error.errorCode.code), error.reason, null); //TODO @BartoszWilk convert this error
+                    }
+                });
+
+        bleAdapter.connectToDevice(deviceId,
+                new ConnectionOptions(isAutoConnect, requestMtu, refreshGattMoment, timeoutMillis, 0 /* BluetoothGatt.CONNECTION_PRIORITY_BALANCED */),
```
pawelByszewski: it would be great to have `0` extracted to a const instead putting the info to a comment

android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java
```diff
@@ -0,0 +1,196 @@
...
+                }, new OnErrorCallback() {
+                    @Override
+                    public void onError(final BleError error) {
+                        safeMainThreadResolver.onError(error);
+                    }
+                });
+    }
+
+    private void observeConnectionState(String deviceId, boolean emitCurrentValue, @NonNull final MethodChannel.Result result) {
+        //emit current value if needed; rest is published automatically through connectToDevice()
```
pawelByszewski: 1. Do we really need a comment? is there any way to write it down using the code?
2. what about javadoc instead of inner comment

android/src/main/java/com/polidea/flutter_ble_lib/event/ConnectionStateStreamHandler.java
```diff
@@ -0,0 +1,25 @@
...
+        this.eventSink = eventSink;
+    }
+
+    @Override
+    public void onCancel(Object o) {
+        eventSink = null;
+    }
+
+    public void onNewConnectionState(ConnectionState newState) {
+        if (eventSink != null) {
```
pawelByszewski: it is not threadsafe, is it ok?

lib/internal/flutter_ble_lib_mixins/device_connection_mixin.dart
```diff
@@ -0,0 +1,60 @@
+part of flutter_ble_lib;
+
+mixin DeviceConnectionMixin on FlutterBLE {
```
pawelByszewski: 👍

android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
pawelByszewski: btw I would like to refactor that :) it is nice entry task :)

android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java
mikolak: 👍

android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java
mikolak: 👍

android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java
mikolak: I don't how to point in code to the fact that a separate and seemingly not connected method publishes most of the stream content. :)

It doesn't concern the user, so I'm not sure if this should be a javadoc.

android/src/main/java/com/polidea/flutter_ble_lib/event/ConnectionStateStreamHandler.java
mikolak: No, good catch!

lib/internal/peripheral_manager.dart
```diff
@@ -0,0 +1,18 @@
+part of flutter_ble_lib;
+
+abstract class PeripheralManager {
```
dariuszseweryn: Shouldn't this be `CentralManager` as the application is in `Central` role?

mikolak: This is an internal interface passed to Peripheral and does not pertain to Central/Peripheral distinction. Perhaps the name should be change to clarify this?

mikolak: I've created an issue for this.

android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
mikolak: Created an issue for this.

PR #101, opened 2019-09-18
Errors handling & scanning feature refactor

bartoszwilk:


android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
```diff
@@ -121,16 +118,12 @@ private void startDeviceScan(@NonNull MethodCall call, Result result) {
...
-                            scanningStreamHandler.onScanResult(data);
-                        } catch (JSONException e) {
-                            e.printStackTrace();
-                        }
+                        scanningStreamHandler.onScanResult(data);
                     }
                 }, new OnErrorCallback() {
                     @Override
                     public void onError(BleError error) {
-                        Log.d("scanning_error", error.toString());
```
pawelByszewski: Imho logging could be useful during development phase, but `Log.e` instead of `Log.d`

android/src/main/java/com/polidea/flutter_ble_lib/converter/BleErrorJsonConverter.java
```diff
@@ -0,0 +1,44 @@
...
+        String CHARACTERISTIC_UUID = "characteristicUUID";
+        String DESCRIPTOR_UUID = "descriptorUUID";
+        String INTERNAL_MESSAGE = "internalMessage";
+    }
+
+    @Override
+    public String toJson(BleError error) throws JSONException {
+        JSONObject root = new JSONObject();
+        root.put(Metadata.ERROR_CODE, error.errorCode.code);
+        if (error.androidCode == null || error.androidCode >= 0x80 || error.androidCode < 0) {
```
pawelByszewski: could you extract `0x80` to some kind of const with descriptive name?

android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
bartoszwilk: I think we should log all errors on MBA side. It can be useful also in RN lib then.

android/src/main/java/com/polidea/flutter_ble_lib/event/ScanningStreamHandler.java
```diff
@@ -22,9 +25,33 @@ public void onCancel(Object o) {
...
-            scanResultsSink.success(scanResultJsonConverter.toJson(scanResult));
+            try {
+                scanResultsSink.success(scanResultJsonConverter.toJson(scanResult));
+            } catch (JSONException e) {
+                e.printStackTrace();
+            }
+        }
+    }
+
+    public void onError(BleError error) {
```
mikolak: All of these methods should be `synchronized` - there could be a race condition on cancelling the stream and pushing new value to it.

android/src/main/java/com/polidea/flutter_ble_lib/converter/BleErrorJsonConverter.java
```diff
@@ -0,0 +1,44 @@
+package com.polidea.flutter_ble_lib.converter;
+
+import com.polidea.multiplatformbleadapter.errors.BleError;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class BleErrorJsonConverter implements JsonConverter<BleError> {
```
mikolak: Could you add the use of this converter in DeviceConnectionDelegate?

lib/error/ble_error.dart
```diff
@@ -0,0 +1,36 @@
+abstract class _Metadata {
```
mikolak: Change the name to `_BleErrorMetadata`. We use only one namespace currently, and using `_Metadata` will create conflics sooner than later.

lib/internal/flutter_ble_lib_mixins/scanning_mixin.dart
```diff
@@ -0,0 +1,27 @@
...
+mixin ScanningMixin on FlutterBLE {
+  final EventChannel _scanEventChannel =
+      const EventChannel(ChannelName.scanningEvents);
+
+  Stream<ScanResult> startDeviceScan(
+    int scanMode,
+    int callbackType,
+    List<String> uuids,
+  ) async* {
+    yield* _scanEventChannel.receiveBroadcastStream().handleError(
```
mikolak: invokeMethod!!11oneone

PR #105, opened 2019-09-20
Package reorganisation and observing connection state of multiple peripherals

mikolak:


android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java
```diff
@@ -97,13 +101,25 @@ public void onError(BleError error) {
...
-                }, new OnEventCallback<ConnectionState>() {
+                },
+                new OnEventCallback<ConnectionState>() {
                     @Override
                     public void onEvent(final ConnectionState data) {
+                        String temp;
+                        try {
+                            temp = new ConnectionStateChangeJsonConverter().toJson(new ConnectionStateChange(deviceId, data));
+                        } catch (JSONException e) {
+                            e.printStackTrace();
```
bartoszwilk: Don't we want to notify dart side about this error (by invoking result.error)?

android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java
```diff
@@ -97,13 +101,25 @@ public void onError(BleError error) {
...
                         safeMainThreadResolver.onSuccess(null);
                     }
-                }, new OnEventCallback<ConnectionState>() {
+                },
+                new OnEventCallback<ConnectionState>() {
                     @Override
                     public void onEvent(final ConnectionState data) {
+                        String temp;
+                        try {
+                            temp = new ConnectionStateChangeJsonConverter().toJson(new ConnectionStateChange(deviceId, data));
```
bartoszwilk: There is no need to create json converter instances on every event. We can declare this converter as local for this method.

android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java
```diff
@@ -115,17 +131,28 @@ public void onError(final BleError error) {
...
 
-    private void observeConnectionState(String deviceId, boolean emitCurrentValue, @NonNull final MethodChannel.Result result) {
+    private void observeConnectionState(final String deviceId, boolean emitCurrentValue, @NonNull final MethodChannel.Result result) {
         //emit current value if needed; rest is published automatically through connectToDevice()
 
         final SafeMainThreadResolver safeMainThreadResolver = new SafeMainThreadResolver<>(
                 new OnSuccessCallback<Boolean>() {
                     @Override
                     public void onSuccess(Boolean data) {
+                        String newData;
```
bartoszwilk: Can we improve naming here? Data says nothing to me (maybe rename it to isConnected). Moreover there is no need to declare "String newData" at the top of the onSuccess callback. We can move it right above the place it is used - try block. We can also rename newData to connectionStateJson or sth like that. What about moving this parsing logic into this particular stream handler?

android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java
```diff
@@ -115,17 +131,28 @@ public void onError(final BleError error) {
...
+                            state = ConnectionState.CONNECTED;
                         else
-                            streamHandler.onNewConnectionState(ConnectionState.DISCONNECTED);
+                            state = ConnectionState.DISCONNECTED;
+
+                        try {
+                            newData = new ConnectionStateChangeJsonConverter().toJson(new ConnectionStateChange(deviceId, state));
+                        } catch (JSONException e) {
+                            e.printStackTrace();
+                            return;
```
bartoszwilk: We should notify dart side about this error.

lib/internal/bridge/device_connection_mixin.dart
```diff
@@ -36,26 +41,23 @@ mixin DeviceConnectionMixin on FlutterBLE {
...
   }
 
-  Future<bool> isPeripheralConnected(
-      String peripheralIdentifier) async {
-    return await _methodChannel
-        .invokeMethod(MethodName.isDeviceConnected, <String, dynamic>{
-      ArgumentName.deviceIdentifier: peripheralIdentifier
-    });
+  Future<bool> isPeripheralConnected(String peripheralIdentifier) async {
+    return await _methodChannel.invokeMethod(MethodName.isDeviceConnected,
```
bartoszwilk: It can be "one-lined"

lib/internal/bridge/device_connection_mixin.dart
```diff
@@ -36,26 +41,23 @@ mixin DeviceConnectionMixin on FlutterBLE {
...
+        <String, dynamic>{ArgumentName.deviceIdentifier: peripheralIdentifier});
   }
 
   Future<void> disconnectOrCancelPeripheralConnection(
       String peripheralIdentifier) async {
-    return await _methodChannel
-        .invokeMethod(MethodName.cancelConnection, <String, dynamic>{
-      ArgumentName.deviceIdentifier: peripheralIdentifier
-    });
+    return await _methodChannel.invokeMethod(MethodName.cancelConnection,
```
bartoszwilk: Can be "one-lined"

android/src/main/java/com/polidea/flutter_ble_lib/event/ConnectionStateStreamHandler.java
```diff
@@ -17,9 +15,9 @@ synchronized public void onCancel(Object o) {
         eventSink = null;
     }
 
-    synchronized public void onNewConnectionState(ConnectionState newState) {
+    synchronized public void onNewConnectionState(String newState) {
```
bartoszwilk: newStateJson?

android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java
```diff
@@ -97,13 +101,25 @@ public void onError(BleError error) {
                     public void onSuccess(Device data) {
                         safeMainThreadResolver.onSuccess(null);
                     }
-                }, new OnEventCallback<ConnectionState>() {
+                },
+                new OnEventCallback<ConnectionState>() {
                     @Override
                     public void onEvent(final ConnectionState data) {
+                        String temp;
```
pawelByszewski: I can imagine a more meaningful name then `temp`:)

lib/internal/bridge/device_connection_mixin.dart
```diff
@@ -18,10 +18,15 @@ mixin DeviceConnectionMixin on FlutterBLE {
...
-    yield* _peripheralConnectionStateEventChannel
-        .receiveBroadcastStream()
-        .map((rawValue) {
-      switch (rawValue) {
+    Stream<PeripheralConnectionState> stream =
+        _peripheralConnectionStateEventChannel
+            .receiveBroadcastStream()
+            .map((jsonString) =>
+                ConnectionStateContainer.fromJson(jsonDecode(jsonString)))
+            .where((container) => container.peripheralIdentifier == identifier)
```
pawelByszewski: `connectionStateContainer` instead of `container`?

lib/internal/bridge/device_connection_mixin.dart
```diff
@@ -18,10 +18,15 @@ mixin DeviceConnectionMixin on FlutterBLE {
...
+            .map((container) => container.connectionState)
```
pawelByszewski: `connectionStateContainer` instead of `container`?

lib/internal/bridge/device_connection_mixin.dart
```diff
@@ -18,10 +18,15 @@ mixin DeviceConnectionMixin on FlutterBLE {
...
+            .map((rawValue) {
```
pawelByszewski: The raw value of what?

lib/internal/bridge/device_connection_mixin.dart
```diff
@@ -36,26 +41,23 @@ mixin DeviceConnectionMixin on FlutterBLE {
       }
     });
 
+    yield* stream;
```
pawelByszewski: it would be great to have a more meaningful name

lib/internal/bridge/device_connection_mixin.dart
mikolak: It's autoformatted.

lib/internal/bridge/device_connection_mixin.dart
mikolak: It's autoformatted.

PR #106, opened 2019-09-23
get log level and fetch log level - dart + android

pawelByszewski:


android/src/main/java/com/polidea/flutter_ble_lib/delegate/CallDelegate.java
```diff
@@ -1,8 +1,21 @@
...
+import java.util.List;
+
 import io.flutter.plugin.common.MethodCall;
 import io.flutter.plugin.common.MethodChannel;
 
-public interface CallDelegate extends MethodChannel.MethodCallHandler {
-    boolean canHandle(MethodCall call);
+abstract public class CallDelegate implements MethodChannel.MethodCallHandler {
+
+    final List<String> supportedMethods;
```
mikolak: 👍

lib/ble_manager.dart
```diff
@@ -2,6 +2,13 @@ part of flutter_ble_lib;
 
 typedef RestoreStateAction = Function(List<Peripheral> restoreStateIdentifier);
 
+enum LogLevel { NONE, VERBOSE, DEBUG, INFO, WARNING, ERROR }
+
+LogLevel logLevelFromString(String logLevelName) {
```
mikolak: Should this be available to user? It looks like an internal function.
`InternalBleManager._logLevelFromString(String logLevelName)`?

lib/ble_manager.dart
```diff
@@ -2,6 +2,13 @@ part of flutter_ble_lib;
 
 typedef RestoreStateAction = Function(List<Peripheral> restoreStateIdentifier);
 
+enum LogLevel { NONE, VERBOSE, DEBUG, INFO, WARNING, ERROR }
```
mikolak: Perhaps `peripheral_connection_state.dart` should be renamed  to `enums.dart` and contain this?

lib/ble_manager.dart
mikolak: If I see correctly it's only used in `log_level_mixin.dart` so it should be there as a private function.

PR #119, opened 2019-09-24
Setup BLEmulator

mikolak:


android/src/main/java/com/polidea/blemulator/SimulatedAdapter.java
```diff
@@ -0,0 +1,262 @@
...
+        return null;
+    }
+
+    @Override
+    public void startDeviceScan(String[] filteredUUIDs,
+                                int scanMode,
+                                int callbackType,
+                                OnEventCallback<ScanResult> onEventCallback,
+                                OnErrorCallback onErrorCallback) {
+        Log.w("SIMULATOR", "START DEVICE SCAN");
```
pawelByszewski: Let's add `log.i` everywhere

android/src/main/java/com/polidea/blemulator/bridging/DartMethodCaller.java
```diff
@@ -0,0 +1,13 @@
...
+
+import io.flutter.plugin.common.MethodChannel;
+
+public class DartMethodCaller {
+    private MethodChannel dartMethodChannel;
+
+    public DartMethodCaller(MethodChannel dartMethodChannel) {
+        this.dartMethodChannel = dartMethodChannel;
+    }
+
```
bartoszwilk: Unnecessary empty lines

android/src/main/java/com/polidea/blemulator/bridging/constants/ChannelName.java
```diff
@@ -0,0 +1,8 @@
+package com.polidea.blemulator.bridging.constants;
+
+public interface ChannelName {
+    String BASE = "com.polidea.blemulator";
+
+    String TO_DART = BASE + "/to_dart";
```
bartoszwilk: We use pascal case for all bridge-related names if I recall correctly.

lib/blemulator/blemulator_core.dart
```diff
@@ -0,0 +1,12 @@
+part of blemulator;
+
+class Blemulator {
+
+  BlemulatorBridge _bridge;
+
+  Blemulator() {
+    _bridge = BlemulatorBridge();
```
bartoszwilk: Why don't we just instantiate this bridge in it's declaration line? We don't inject it (maybe we should?) so currently there is no need to have this default constructor.

PR #135, opened 2019-09-25
call MTU result on main thread; show negotiated MTU in sample

pawelByszewski:


android/src/main/java/com/polidea/flutter_ble_lib/delegate/MtuDelegate.java
```diff
@@ -48,15 +49,31 @@ public void onMethodCall(@NonNull MethodCall methodCall, @NonNull MethodChannel.
 
     private void requestMtu(String deviceIdentifier, @NonNull int mtu, String transactionId, @NonNull final MethodChannel.Result result) {
         Log.d(TAG, "Request MTU " + mtu);
+
+        final SafeMainThreadResolver resolver = new SafeMainThreadResolver<>(
```
mikolak: Does RSSI need this as well?

PR #137, opened 2019-09-26
Refactor test scenario

mikolak:


example/lib/device_details/device_details_bloc.dart
```diff
@@ -77,175 +77,32 @@ class DeviceDetailsBloc {
...
-    Function log = (text) {
+    Logger log = (text) {
       logs.insert(0, DebugLog(DateTime.now().toString(), text));
+      Fimber.d(text);
+      _logsController.add(logs);
+    };
+
+    Logger logError = (text) {
+      logs.insert(0, DebugLog(DateTime.now().toString(), text.toUpperCase()));
+      Fimber.d(text.toUpperCase());
```
pawelByszewski: 1. `Fimber.e`?
2. it is almost the same as `log` (except double `toUpperCase`) so let's just  change letters in text and call another function instead of copying the logic

example/lib/device_details/device_details_bloc.dart
```diff
@@ -77,175 +77,32 @@ class DeviceDetailsBloc {
...
       _logsController.add(logs);
     };
 
+    log("Setting log level to debug");
+    await _bleManager.setLogLevel(LogLevel.debug);
```
pawelByszewski: try to fetch `logLevel` to check if the previous line works correctly

example/lib/test_scenario_view.dart
```diff
@@ -113,7 +113,7 @@ class _TestScenarioWidgetState extends State<TestScenarioWidget> {
                     child: Text("Run Sensor Tag test"),
                     onPressed: _isTestInProgress
                         ? null
-                        : () => _startTestScenario(SensorTagTestScenario()),
+                        : () => _startTestScenario(SensorTagTestWithScanAndConnectionScenario()),
```
pawelByszewski: is that view displayed in the app?

example/lib/test_scenarios/sensor_tag_test_scenario.dart
```diff
@@ -1,97 +1,195 @@
...
-        .then((_) => log("\BleClient destroyed after a delay"));
+    log("PRINTING CHARACTERISTICS FROM \nPERIPHERAL for the same service");
+    List<Characteristic> characteristicFromPeripheral =
+        await _peripheral.characteristics(service.uuid);
+    characteristicFromPeripheral.forEach((characteristic) =>
+        log("Found characteristic \n ${characteristic.uuid}"));
+    return peripheral;
+  }
+
+  Future<Peripheral> _rssi(
```
pawelByszewski: could you change the method name to something more meaningful? a verb would be helpful

example/lib/test_scenarios/sensor_tag_test_scenario.dart
```diff
@@ -1,97 +1,195 @@
...
+      Logger log, Logger logError, Peripheral peripheral) async {
+    int rssi = await peripheral.rssi();
+    log("rssi $rssi");
+    return peripheral;
+  }
+
+  Future<Peripheral> _mtu(
```
pawelByszewski: could you change the method name to something more meaningful? a verb would be helpful

example/lib/test_scenarios/sensor_tag_test_scenario.dart
```diff
@@ -1,97 +1,195 @@
...
 
-class SensorTagTestScenario implements TestScenario {
-  BleManager bleManager = BleManager.getInstance();
-  bool deviceConnectionAttempted = false;
+class SensorTagTestScenario extends TestScenario {
+  final Peripheral _peripheral;
 
+  SensorTagTestScenario(this._peripheral);
+
+  @override
```
pawelByszewski: what about pass loggers as the constructor parameters? then we wouldn't have to pass it to every private methods?

PR #139, opened 2019-09-26
Characteristics monitoring

bartoszwilk:


bartoszwilk:
@mikolak 
Ad.1 Done.
Ad.2 Done.

pawelByszewski:
maybe we should discuss it at first if it makes sense to do the rebase? that way this PR(that is generally ok) is blocked by another one.

pawelByszewski:
Conflicts, conflicts everywhere

example/lib/test_scenarios/sensor_tag_test_scenario.dart
```diff
@@ -74,7 +75,17 @@ class SensorTagTestScenario {
     CharacteristicWithValue readValue = await peripheral.readCharacteristic(
         SensorTagTemperatureUuids.temperatureService,
         SensorTagTemperatureUuids.temperatureDataCharacteristic);
-    log("Reada temperature value ${readValue.value}");
+    log("Read temperature value ${readValue.value}");
+
+    log("Start monitoring temperature");
```
mikolak: Move monitoring before the first write and change name of the method to `_readWriteMonitor`

example/lib/test_scenarios/sensor_tag_test_scenario.dart
```diff
@@ -120,7 +131,17 @@ class SensorTagTestScenario {
     CharacteristicWithValue dataCharacteristic =
         await service.readCharacteristic(
             SensorTagTemperatureUuids.temperatureDataCharacteristic);
-    log("Read temperature value ${dataCharacteristic.value}");
+    log("Read temperature value ${_convertToTemperature(dataCharacteristic.value)}C");
+
+    log("Start monitoring temperature");
```
mikolak: Move before first write and change name to `_readWriteMonitor...`

example/lib/test_scenarios/sensor_tag_test_scenario.dart
```diff
@@ -161,15 +184,23 @@ class SensorTagTestScenario {
 
     log("Reading characteristic value");
     Uint8List value = await characteristic.read();
-    log("Read temperature config value $value");
+    log("Read temperature config value ${_convertToTemperature(value)}C");
+
+    log("Start monitoring temperature");
```
mikolak: Move before first write, change method name

example/lib/test_scenarios/sensor_tag_test_scenario.dart
```diff
@@ -179,4 +210,28 @@ class SensorTagTestScenario {
...
   }
+
+  void _startMonitoringTemperature(
+      Stream<Uint8List> characteristicUpdates, Function log) {
+    log("START TEMPERATURE MONITORING");
+    monitoringStreamSubscription?.cancel();
+    monitoringStreamSubscription = characteristicUpdates
+        .map(_convertToTemperature)
+        .where((temperature) => temperature != 0)
+        .take(1)
```
mikolak: Why take(1)? This method should return stream subscription which should be disposed at the end of the test method

bartoszwilk: It's just for the sake of logs clarity. The update interval is really fast and it shows duplicated values in a really short period.Even if you turn the update off, the temperature characteristic will be notified about update with temperature value set to 0 (it will be done multiple times in a fast way). Take(1) is to prevent from spamming in our logs

PR #142, opened 2019-09-27
Check & observe radio state

bartoszwilk:


example/lib/test_scenarios/bluetooth_state_toggle_scenario.dart
```diff
@@ -22,6 +29,8 @@ class BluetoothStateTestScenario implements TestScenario {
 
     log("Enabled radio!");
 
+    log("Get radio state: ${bleManager.state()}");
```
mikolak: Won't this log `Future...` instead of actual state? Shouldn't there be an await?

example/lib/test_scenarios/bluetooth_state_toggle_scenario.dart
```diff
@@ -13,6 +15,11 @@ class BluetoothStateTestScenario implements TestScenario {
     await bleManager.createClient();
     log("Created client");
 
+    log("Subscribe for radio state changes");
+    _observeRadioState(bleManager, log);
+
+    log("Get radio state: ${bleManager.state()}");
```
mikolak: Won't this log Future... instead of actual state? Shouldn't there be an await?

example/lib/test_scenarios/bluetooth_state_toggle_scenario.dart
```diff
@@ -30,7 +39,17 @@ class BluetoothStateTestScenario implements TestScenario {
     log("Disabled radio!");
 
     log("Destroying client");
+    await _radioStateSubscription?.cancel();
```
mikolak: Why await?

lib/internal/bridge/bluetooth_state_mixin.dart
```diff
@@ -14,4 +17,28 @@ mixin BluetoothStateMixin on FlutterBLE {
...
+    await _methodChannel.invokeMethod(MethodName.observeState);
+    yield* _adapterStateChanges
+        .receiveBroadcastStream()
+        .map(_mapToBluetoothState);
+  }
+
+  BluetoothState _mapToBluetoothState(dynamic rawValue) {
+    switch (rawValue) {
+      case "PoweredOn":
+        return BluetoothState.POWERED_ON;
```
mikolak: Aren't there also `TURNING_ON`, `TURNING_OFF`?

example/lib/test_scenarios/bluetooth_state_toggle_scenario.dart
bartoszwilk: Cancel returns future and I wanted to be sure that this operation will be completed before subscribing for the stream again

lib/internal/bridge/bluetooth_state_mixin.dart
mikolak: https://developer.apple.com/documentation/corebluetooth/cbmanagerstate

There are, but not on iOS and we need to conform

lib/ble_manager.dart
```diff
@@ -37,4 +37,8 @@ abstract class BleManager {
   Future<void> enableRadio({String transactionId});
 
   Future<void> disableRadio({String transactionId});
+
+  Future<BluetoothState> state();
+
+  Stream<BluetoothState> onStateChange();
```
mikolak: `Stream<BluetoothState> observeState({bool emitCurrentValue = true});`?

lib/internal/bridge/bluetooth_state_mixin.dart
bartoszwilk: Yes there are, abut apparently iOS doesn't support them so we introduced resetting state. I also added unknown, unsupported and unauthorized states as it is in RN lib.

lib/ble_manager.dart
bartoszwilk: Fixed

PR #143, opened 2019-09-27
BLEmulator: scanning for peripherals

mikolak:
Blocked by #138 

Still has to be refactored


android/src/main/java/com/polidea/blemulator/bridging/DartMethodCaller.java
```diff
@@ -8,4 +15,90 @@
...
     }
+
+    public void createClient() {
+        HashMap<String, Object> arguments = new HashMap<>();
+        dartMethodChannel.invokeMethod(UpstreamMethodName.CREATE_CLIENT,
+                arguments,
+                new MethodChannel.Result() {
+                    @Override
+                    public void success(@Nullable Object o) {
+                        Log.e("BLEMULATOR BRIDGE", "CREATE CLIENT SUCCESS");
```
pawelByszewski: 1. should be it logged on error level?
2. `BLEMULATOR BRIDGE` export to const

android/src/main/java/com/polidea/blemulator/bridging/DartMethodCaller.java
```diff
@@ -8,4 +15,90 @@
...
+                    }
+
+                    @Override
+                    public void error(String s, @Nullable String s1, @Nullable Object o) {
+
```
pawelByszewski: 1. let's log something in every empty method

lib/blemulator/example/example_peripheral.dart
```diff
@@ -0,0 +1,53 @@
...
+
+class SensorTag extends SimulatedPeripheral {
+  SensorTag()
+      : super("SensorTag", "4B:99:4C:34:DE:77", Duration(milliseconds: 500), [
+          SimulatedService(
+              "F000AA00-0451-4000-B000-000000000000",
+              true,
+              [
+                SimulatedCharacteristic(
+                    "12:25:00:00:00:00:00:00:00:00:B0:00:40:51:04:01:AA:00:F0",
```
pawelByszewski: what about using named parameters for every argument?

lib/blemulator/simulation/simulated_peripheral.dart
```diff
@@ -1,22 +1,67 @@
...
+  Uint8List manufacturerData;
+  Map<String, Uint8List> serviceData;
+  List<String> serviceUuids;
+
+  String localName;
+  List<String> solicitedServiceUuids;
+  List<String> overflowUuids;
+
+  ScanInfo(
+      {this.rssi = -30,
```
pawelByszewski: let's extract default values to some config or something like that

lib/blemulator/example/example_peripheral.dart
mikolak: Dart supports it only for optional arguments and these should be required. :(

android/src/main/java/com/polidea/blemulator/bridging/DartValueHandler.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    public void setScanResultErrorPublisher(OnErrorCallback scanResultErrorPublisher) {
+        this.scanResultErrorPublisher = scanResultErrorPublisher;
+    }
+
+    @Override
+    public void onMethodCall(@NonNull MethodCall call, @NonNull MethodChannel.Result result) {
+        switch (call.method) {
+            case DownstreamMethodName.ADD_SCAN_RESULT:
```
bartoszwilk: Maybe emitScanResult? publishScanResult?

lib/blemulator/example/example_peripheral.dart
bartoszwilk: There is already @required annotation in Dart, but it just causes lint warnings instead of compilation error.
  void test(int id, {@required int param}) {}

lib/blemulator/internal/bridge/upstream_bridge.dart
```diff
@@ -0,0 +1,43 @@
...
+        return _startDeviceScan();
+      case UpstreamMethodName.stopDeviceScan:
+        return _stopDeviceScan();
+      default:
+        throw UnimplementedError("${call.method} is not implemented");
+    }
+  }
+
+  Future<void> _createClient() async {
+    _manager._createClient();
```
bartoszwilk: Shouldn't we "await" for this method to complete?

android/src/main/java/com/polidea/blemulator/SimulatorPlugin.java
```diff
@@ -27,16 +29,17 @@ public static void registerWith(PluginRegistry.Registrar registrar) {
...
     private SimulatorPlugin(MethodChannel upstream) {
         dartMethodCaller = new DartMethodCaller(upstream);
+        dartValueHandler = new DartValueHandler();
     }
 
     @Override
     public void onMethodCall(@NonNull MethodCall call, @NonNull MethodChannel.Result result) {
         switch (call.method) {
-            case MethodName.SIMULATE:
+            case DownstreamMethodName.SIMULATE:
```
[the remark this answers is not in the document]
mikolak: I'll change it to `PlatformMethodName` and the upstream to `DartMethodName`. I'd like to keep names of both on all 3 sides, so I'll avoid Java/Android in the name.

android/src/main/java/com/polidea/blemulator/SimulatorPlugin.java
```diff
@@ -17,6 +18,7 @@
 public class SimulatorPlugin implements MethodChannel.MethodCallHandler {
 
     private DartMethodCaller dartMethodCaller;
+    private DartValueHandler dartValueHandler;
 
     public static void registerWith(PluginRegistry.Registrar registrar) {
         MethodChannel downstream = new MethodChannel(registrar.messenger(), ChannelName.TO_JAVA);
```
bartoszwilk: Just a suggestion. Maybe we should get rid of all downstream/upstream conventions since we partially did so?

lib/blemulator/internal/bridge/constants.dart
```diff
@@ -6,6 +6,33 @@ abstract class ChannelName {
...
 }
 
-abstract class MethodName {
+abstract class DownstreamMethodName {
   static String simulate = "simulate";
-}
\ No newline at end of file
+}
+
+abstract class SimulationDownstreamMethodName {
```
bartoszwilk: Just a suggestion. Maybe we should get rid of all downstream/upstream names since we partially did so?

lib/blemulator/internal/bridge/constants.dart
```diff
@@ -6,6 +6,33 @@ abstract class ChannelName {
...
+  static const String publishScanResult = "publishScanResult";
+}
+
+abstract class UpstreamMethodName {
```
bartoszwilk: Just a suggestion. Maybe we should get rid of all downstream/upstream names since we partially did so?

lib/blemulator/internal/bridge/constants.dart
```diff
@@ -6,6 +6,33 @@ abstract class ChannelName {
   static String downstream = "$_base/toJava";
 }
 
-abstract class MethodName {
+abstract class DownstreamMethodName {
```
bartoszwilk: PlatformMethodName?

android/src/main/java/com/polidea/blemulator/bridging/constants/PlatformMethodName.java
```diff
@@ -0,0 +1,11 @@
+package com.polidea.blemulator.bridging.constants;
+
+public interface PlatformMethodName {
+    String SIMULATE = "simulate";
+
+    String SET_ADAPTER_STATE = "setAdapterState";
+    String PUBLISH_SCAN_RESULT = "publishScanResult";
+    String SET_CONNECTION_STATE = "";
```
pawelByszewski: is it ok?

mikolak: Fixed, good catch

PR #158, opened 2019-09-30
Feature/[ios] Connect to device

TomoLV:


ios/Classes/Util/ResultConverter.m
```diff
@@ -1,8 +1,8 @@
-#import "ScanResultFactory.h"
+#import "ResultConverter.h"
 
-@implementation ScanResultFactory
+@implementation ResultConverter
```
mikolak: It's not aware of any result, it just gets a serializable object. `JSONStringifier`?

TomoLV: Changed to JSONStringifier

PR #166, opened 2019-10-02
Fix/[ios] BleError  handling capabilities

TomoLV:


ios/Classes/Event/ScanningStreamHandler.m
```diff
@@ -27,7 +27,7 @@ - (void)onScanResult:(NSArray *)scanResult {
             if (scanResult[0] == [NSNull null]) {
                 scanResultsSink([JSONStringifier jsonStringFromJSONObject:scanResult[1]]);
             } else {
-                scanResultsSink([FlutterErrorFactory flutterErrorFromJSONString:scanResult[0]]);
+                scanResultsSink([FlutterErrorFactory flutterErrorForJSONString:scanResult[0]]);
```
mikolak: ?

TomoLV: Method was renamed in FlutterErrorFactory so it had to be changed here.

TomoLV: There was also `flutterErrorForBleError` method added so change from `from` to `for` in the method name somehow sounded better for me.

mikolak: But we want to create a Flutter-understandable error _from_ an instance of BleError, right?

mikolak:
In conversation with @TomoLV it turned out these changes are unnecessary.

PR #177, opened 2019-10-03
Simulated connecting, disconnecting, observing and checking connection state

### LINKED ISSUE #148: Connect to Device [1]
* a client is able to connect to a device
* `isConnected`
* observe connection state
* disconnect or cancel connection

mikolak:
Resolves #148


lib/blemulator/internal/bridge/dart_to_platform_bridge.dart
```diff
@@ -1,29 +1,54 @@
...
+          SimulationPlatformMethodName.publishConnectionState,
+          <String, dynamic>{
+            SimulationArgumentName.id: peripheral.id,
+            SimulationArgumentName.connectionState: _connectionStateToString(connectionState),
+          });
+
+  String _connectionStateToString(PeripheralConnectionState state) {
+    switch (state) {
+      case PeripheralConnectionState.connecting:
+        return "CONNECTING";
```
bartoszwilk: We already have NativeConnectionState class for these values.

lib/blemulator/internal/bridge/platform_to_dart_bridge.dart
```diff
@@ -1,43 +1,62 @@
...
     await _manager._startDeviceScan();
   }
 
-  Future<void> _stopDeviceScan() async {
+  Future<void> _stopDeviceScan(MethodCall call) async {
     await _manager._stopDeviceScan();
   }
+
+  Future<void> _connectToDevice(MethodCall call) async {
+    await _manager._connectToDevice(call.arguments[ArgumentName.id] as String);
```
bartoszwilk: Why can't we just return the value from manager's connectToDevice method? I think there is no need for making this methods async and performing await here since final await is being done in _handleAsMethodCall in platform_channel.dart

lib/blemulator/internal/simulation_manager.dart
```diff
@@ -23,18 +30,79 @@ class SimulationManager {
...
+      if (canConnect) {
+        _peripherals[identifier].onConnect();
+      } else {
+        return Future.error(<String, dynamic>{
+          "errorCode": BleErrorCode.DeviceAlreadyConnected,
+          "reason": "Peripheral $identifier is already connected",
+        });
+      }
+    } else {
+      return Future.error(<String, dynamic>{
```
bartoszwilk: We can react for this case at the top of this function and then we will not need this external if.

lib/blemulator/internal/simulation_manager.dart
```diff
@@ -23,18 +30,79 @@ class SimulationManager {
...
+            }
+          }));
+
+      bool canConnect = await _peripherals[identifier].onConnectRequest();
+
+      if (canConnect) {
+        _peripherals[identifier].onConnect();
+      } else {
+        return Future.error(<String, dynamic>{
+          "errorCode": BleErrorCode.DeviceAlreadyConnected,
```
bartoszwilk: I think that this information should be gotten from peripheral, because now we hardcoded just a one reason why we cannot connect with selected peripheral.

bartoszwilk: Maybe onConnectRequest should throw an error? Or maybe onConnect should determine how the connection process should look like and then we could just return _peripherals[identifier].onConnect from here?

lib/blemulator/internal/simulation_manager.dart
```diff
@@ -23,18 +30,79 @@ class SimulationManager {
...
+            if (connectionState == PeripheralConnectionState.disconnected) {
+              _connectionStateSubscriptions[identifier].cancel();
+              _connectionStateSubscriptions.remove(identifier);
+            }
+          }));
+
+      bool canConnect = await _peripherals[identifier].onConnectRequest();
+
+      if (canConnect) {
+        _peripherals[identifier].onConnect();
```
bartoszwilk: Are we missing a "return" here?

lib/blemulator/internal/simulation_manager.dart
```diff
@@ -23,18 +30,79 @@ class SimulationManager {
...
+        "reason": "Unknown peripheral identifier $identifier",
+      });
+    }
+  }
+
+  Future<bool> _isDeviceConnected(String identifier) async {
+    if (_peripherals[identifier] != null) {
+      return _peripherals[identifier].isConnected();
+    } else {
+      return Future.error(<String, dynamic>{
```
bartoszwilk: We could extract this code to separate method as it is used in more places.

lib/blemulator/internal/simulation_manager.dart
```diff
@@ -23,18 +30,79 @@ class SimulationManager {
...
+    if (_peripherals[identifier] != null) {
+      if (!_peripherals[identifier].isConnected()) {
+        return Future.error(<String, dynamic>{
+          "errorCode": BleErrorCode.DeviceDisconnected,
+          "reason": "Peripheral $identifier already disconnected",
+        });
+      }
+      return _peripherals[identifier].onDisconnect();
+    } else {
+      return Future.error(<String, dynamic>{
```
bartoszwilk: We could extract this code to separate method as it is used in more places.

lib/blemulator/internal/simulation_manager.dart
```diff
@@ -23,18 +30,79 @@ class SimulationManager {
...
+        "errorCode": BleErrorCode.DeviceNotFound,
+        "reason": "Unknown peripheral identifier $identifier",
+      });
+    }
+  }
+
+  Future<void> _disconnectOrCancelConnection(String identifier) async {
+    if (_peripherals[identifier] != null) {
```
bartoszwilk: Same as above. We can invoke one function here that checks if device is known and return error if not.

lib/blemulator/internal/simulation_manager_mixins/error_checks_mixin.dart
```diff
@@ -0,0 +1,39 @@
...
+    if (_peripherals[identifier].isConnected()) {
+      return Future.error(<String, dynamic>{
+        "errorCode": BleErrorCode.DeviceAlreadyConnected,
+        "reason": "Peripheral $identifier is already connected",
+      });
+    }
+  }
+
+  Future<void> _errorIfNotConnected(String identifier) async {
+    if (_peripherals[identifier] == null) {
```
bartoszwilk: We should check if device is connected here

lib/blemulator/internal/simulation_manager_mixins/error_checks_mixin.dart
```diff
@@ -0,0 +1,39 @@
...
+      return Future.error(<String, dynamic>{
+        "errorCode": BleErrorCode.DeviceNotConnected,
+        "reason": "Peripheral $identifier is not connected",
+      });
+    }
+  }
+
+  Future<void> _errorIfUnknown(String identifier) async {
+    if (!_peripherals[identifier].isConnected()) {
```
bartoszwilk: We should check if device is null here

lib/blemulator/internal/simulation_manager_mixins/client_managing_mixin.dart
```diff
@@ -0,0 +1,9 @@
+part of internal;
+
+mixin ClientManagingMixin on SimulationManagerBase {
+  Future<void> _createClient() async {
+    print("entered create client");
+  }
+
+  Future<void> _destroyClient() async {}
```
bartoszwilk: Don't we want to log that this method was invoked?

lib/blemulator/internal/simulation_manager_mixins/peripheral_scanning_mixin.dart
```diff
@@ -0,0 +1,20 @@
...
+
+mixin PeripheralScanningMixing on SimulationManagerBase {
+  List<StreamSubscription> _scanSubscriptions = [];
+
+  Future<void> _startDeviceScan() async {
+    _peripherals.values.forEach((peripheral) {
+      _scanSubscriptions
+          .add(peripheral.onScan(allowDuplicates: true).listen((scanResult) {
+        print(scanResult);
+        return _bridge.publishScanResult(scanResult);
```
bartoszwilk: Can we swap "return" for "await" here? It would be much more intuitive.

android/src/main/java/com/polidea/blemulator/bridging/DartMethodCaller.java
```diff
@@ -103,4 +113,94 @@ public void notImplemented() {
...
+        dartMethodChannel.invokeMethod(DartMethodName.CONNECT_TO_DEVICE, arguments, new MethodChannel.Result() {
+            @Override
+            public void success(@Nullable Object o) {
+                Log.d(TAG, "connectToDevice SUCCESS");
+                onSuccessCallback.onSuccess(new Device(deviceIdentifier, name));
+            }
+
+            @Override
+            public void error(String s, @Nullable String s1, @Nullable Object o) {
+                Log.e(TAG, s);
```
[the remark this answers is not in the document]
mikolak: I agree. Here's the issue: https://github.com/Polidea/FlutterBleLib/issues/182

android/src/main/java/com/polidea/blemulator/bridging/DartMethodCaller.java
```diff
@@ -103,4 +113,94 @@ public void notImplemented() {
...
+            @Override
+            public void success(@Nullable Object o) {
+                Log.d(TAG, "connectToDevice SUCCESS");
+                onSuccessCallback.onSuccess((Boolean) o);
+            }
+
+            @Override
+            public void error(String s, @Nullable String s1, @Nullable Object o) {
+                Log.e(TAG, s);
+                onErrorCallback.onError(objectToBleError(o));
```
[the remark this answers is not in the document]
mikolak: I've created an issue for this: https://github.com/Polidea/FlutterBleLib/issues/182

android/src/main/java/com/polidea/blemulator/bridging/DartMethodCaller.java
```diff
@@ -103,4 +113,94 @@ public void notImplemented() {
...
+
+            @Override
+            public void notImplemented() {
+                Log.e(TAG, "CONNECT TO DEVICE not implemented");
+            }
+        });
+    }
+
+    private BleError objectToBleError(Object o) {
+        Map<String, Object> error = (Map<String, Object>) o;
```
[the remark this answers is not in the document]
mikolak: I've created an issue for this: https://github.com/Polidea/FlutterBleLib/issues/182

example/lib/test_scenarios/sensor_tag_test_scenario.dart
```diff
@@ -12,15 +12,15 @@ class SensorTagTestScenario {
...
-        .then((_) => _cancelTransaction())
-        .then((_) => _discovery())
-        .then((_) => _testRequestingMtu())
-        .then((_) => _testReadingRssi())
-        .then((_) => _readWriteMonitorCharacteristicForPeripheral())
-        .then((_) => _readWriteMonitorCharacteristicForService())
-        .then((_) => _readWriteMonitorCharacteristic())
-        .then((_) => _fetchConnectedDevice())
-        .then((_) => _fetchKnownDevice())
+//        .then((_) => _cancelTransaction())
```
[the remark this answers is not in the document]
mikolak: None of these test cases have been implemented in the BLEmulator yet, but we can test disconnecting, which is the last one used. They will be uncommented one by one as we move forward with features.

example/lib/device_details/device_details_bloc.dart
```diff
@@ -86,17 +86,17 @@ class DeviceDetailsBloc {
...
-    log("Current log level $logLevel");
-
-    log("Setting log level to debug");
-    await _bleManager.setLogLevel(LogLevel.debug);
-    log("Set log level to debug");
-
-    log("Fetching log level");
-    logLevel = await _bleManager.logLevel();
-    log("Current log level $logLevel");
+//    log("Fetching log level");
```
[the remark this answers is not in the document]
mikolak: Yep, good catch 👍

lib/blemulator/internal/bridge/dart_to_platform_bridge.dart
mikolak: We shouldn't use internals of FlutterBleLib inside BLEmulator. I've created a separate definition for those values in this module.

lib/blemulator/internal/simulation_manager.dart
mikolak: 1) I made a mistake with conditions, it's fixed now.
2) Ad `onConnectRequest()` I think this gives user more control over the behaviour of the simulated peripheral. I think there are some devices that would not even attempt connection and would just reject any such requests.

lib/blemulator/internal/simulation_manager_mixins/client_managing_mixin.dart
mikolak: It's logged from Java.

PR #186, opened 2019-10-04
[iOS] Control, read, monitor BT State

TomoLV:


mikolak:
@bartoszwilk will push out a modified sample to test this

TomoLV:
Tested getState with the expected outcome.
Enable and disable radio methods are not supported on iOS and they result in an error as expected.

PR #187, opened 2019-10-04
Simulated characteristics reading

bartoszwilk:


lib/blemulator/internal/bridge/platform_to_dart_bridge.dart
```diff
@@ -40,4 +57,56 @@ class PlatformToDartBridge {
...
+        )
+        .then((characteristic) => _convertToMap(
+            arguments[ArgumentName.deviceIdentifier], characteristic));
+  }
+
+  Map<String, dynamic> _convertToMap(String peripheralId, characteristic) =>
+      <String, dynamic>{
+        ArgumentName.deviceIdentifier: peripheralId,
+        ArgumentName.characteristicUuid: characteristic.uuid,
+        //TODO Change to SimulationArgumentName after merge with peripheral connection feature
```
mikolak: Now's the time!

lib/blemulator/internal/simulation_manager.dart
```diff
@@ -24,17 +24,92 @@ class SimulationManager {
...
     _scanSubscriptions.forEach((subscription) => subscription.cancel());
     _scanSubscriptions.clear();
   }
+
+  Future<SimulatedCharacteristic> _readCharacteristicForIdentifier(
+    int characteristicIdentifier,
+  ) {
+    SimulatedCharacteristic targetCharacteristic;
+    peripheralsLoop:
+    for (SimulatedPeripheral peripheral in _peripherals) {
```
mikolak: 1) I propose `SimulationManager.getCharacteristic(int id)`.
2) I propose `SimulatedPeripheral.getCharacteristic(int id)`.

Former extracts outer loop for reuse in write/monitor, latter extracts inner loop for the same purposes.

lib/blemulator/internal/simulation_manager.dart
```diff
@@ -24,17 +24,92 @@ class SimulationManager {
...
+        ?.characteristics()
+        ?.firstWhere(
+          (characteristic) => characteristic.uuid == characteristicUUID,
+          orElse: () => null,
+        );
+
+    if (targetCharacteristic == null)
+      return Future.error("Characteristic not found");
+
+    return targetCharacteristic.read().then(
```
mikolak:
```
await targetCharacteristic.read();
return targetCharacteristic;
```
or
```
SimulatedCharacteristic SimulatedCharacteristic.read();
```
?

mikolak: Actually, it should be `SimulatedCharacteristic._value` and since that should be private we should have some intermediate representation, like `CharacteristicResponse(Characteristic, Uint8List)`.

lib/blemulator/simulation/simulated_characteristic.dart
```diff
@@ -3,19 +3,32 @@ part of blemulator;
...
-      : id = IdGenerator().nextId();
+  SimulatedCharacteristic({
+    @required this.uuid,
+    @required this.value,
+    this.convenienceName,
+    this.isReadable = true,
+    this.isWritableWithResponse = true,
+    this.isWritableWithoutResponse = true,
+    this.isNotifiable = false,
+    this.isIndicatable = false,
```
mikolak: There should also be `isNotifying`

lib/blemulator/internal/simulation_manager.dart
bartoszwilk: Extracted.

bartoszwilk:
Merged in PR https://github.com/Polidea/FlutterBleLib/pull/202

PR #190, opened 2019-10-07
[iOS] Handle disconnected event

TomoLV:


ios/Classes/Event/ConnectionStateStreamHandler.m
```diff
@@ -26,6 +27,18 @@ - (void)onConnectedEvent:(NSString *)deviceId {
     }
 }
 
+- (void)onDisconnectedEvent:(NSArray *)peripheralResponse {
```
mikolak: Is this thread safe?

TomoLV: Added thread safe handling.

PR #191, opened 2019-10-07
[iOS] Observe connection state

TomoLV:


ios/Classes/FlutterBleLibPlugin.m
```diff
@@ -167,6 +169,16 @@ - (void)isDeviceConnected:(FlutterMethodCall *)call result:(FlutterResult)result
 
 }
 
+- (void)observeConnectionState:(FlutterMethodCall *)call result:(FlutterResult)result {
+    BOOL emitCurrentValue = call.arguments[ARGUMENT_KEY_EMIT_CURRENT_VALUE];
+    if (emitCurrentValue == YES) {
+        [self isDeviceConnected:call result:result];
```
mikolak: The current value should be emitted to the stream, not returned in the result.

TomoLV: Pushed changes to address this issue.

PR #197, opened 2019-10-07
[iOS] Add constants

TomoLV:


ios/Classes/Blemulator/Bridging/Constants/SimulationChannelName.m
```diff
@@ -0,0 +1,4 @@
+#define BASE @"com.polidea.blemulator"
+
+NSString * const SIMULATION_CHANNEL_NAME_TO_DART = (BASE @"/toDart");
+NSString * const SIMULATION_CHANNEL_NAME_TO_OBJC = (BASE @"/toObjC");
```
mikolak: Nu-uh, it's the same channel as the one used in Java, so either `/toJava` or rename both to `/toPlatform`. The latter sounds better.

TomoLV: I like `/toPlatform`

TomoLV:
Closing this PR.

Constants are going to be added feature by feature whenever they are needed.

PR #198, opened 2019-10-07
Add iOS specific argument to startPeripheralScan

### LINKED ISSUE #134: Add allowDuplicates argument to startDeviceScan() in Flutter for iOS [0.125]
iOS Library uses optional `allowDuplicates` argument in `startDeviceScan(_ filteredUUIDs: [String]?, options:[String:AnyObject]?)` as one of the `options`.

```
// iOS handles allowDuplicates option to receive more scan records.
        var rxOptions = [String:Any]()
        if let options = options {
            if ((options["allowDuplicates"]?.isEqual(to: NSNumber(value: true as Bool))) ?? false) {
                rxOptions[CBCentralManagerScanOptionAllowDuplicatesKey] = true
            }
        }
```

Currently Flutter example app calls startDevieScan without this argument at all.

mikolak:
Fixes #134 
@TomoLV please confirm it's working

### DIFF
--- a/ios/Classes/Constants/ArgumentKey.h
+++ b/ios/Classes/Constants/ArgumentKey.h
@@ -2,6 +2,7 @@
 
 extern NSString * const ARGUMENT_KEY_RESTORE_STATE_IDENTIFIER;
 extern NSString * const ARGUMENT_KEY_SCAN_MODE;
+extern NSString * const ARGUMENT_KEY_ALLOW_DUPLICATES;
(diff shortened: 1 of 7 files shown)


TomoLV:
There was some work needed on the iOS part to handle the argument. I've pushed one commit containing required adjustments.
Tested, it's working properly, the argument is passed from dart to objc properly.

Ready to merge from my perspective.

PR #201, opened 2019-10-08
Feature/example refactor

pawelByszewski:


example/lib/device_details/device_details_bloc.dart
```diff
@@ -39,27 +43,81 @@ class DeviceDetailsBloc {
...
     if (await _deviceController.stream.value.peripheral.isConnected()) {
+      log("DISCONNECTING...");
       await _deviceController.stream.value.peripheral
           .disconnectOrCancelConnection();
     }
-    return _deviceRepository.pickDevice(null);
+    log("Disconnected!");
+  }
+
+  Future<void> readRssi() async {
```
bartoszwilk: There is no need to make it async since we don't "await" for anything.

example/lib/device_details/device_details_bloc.dart
```diff
@@ -39,27 +43,81 @@ class DeviceDetailsBloc {
...
+    _clearLogs();
+    _deviceController.stream.listen((bleDevice) async {
```
bartoszwilk: There is no need to make it async since we don't "await" for anything.

example/lib/test_scenarios/peripheral_test_operations.dart
```diff
@@ -0,0 +1,269 @@
...
+  final BleManager bleManager;
+
+  PeripheralTestOperations(
+      this.bleManager, this.peripheral, this.log, this.logError);
+
+  Future<void> connect() async {
+    log("Connecting to ${peripheral.name}");
+    await peripheral.connect();
+    log("Connected!");
+    return peripheral;
```
bartoszwilk: What is the purpose of returning peripheral here since this method should return Future<'void'>?

example/lib/test_scenarios/peripheral_test_operations.dart
```diff
@@ -0,0 +1,269 @@
...
+    await peripheral.characteristics(service.uuid);
+    characteristicFromPeripheral.forEach((characteristic) =>
+        log("Found characteristic \n ${characteristic.uuid}"));
+    return peripheral;
+  }
+
+  Future<void> testReadingRssi() async {
+    int rssi = await peripheral.rssi();
+    log("rssi $rssi");
+    return peripheral;
```
bartoszwilk: Same as above

example/lib/test_scenarios/sensor_tag_test_scenario.dart
```diff
@@ -1,279 +1,28 @@
...
-        .connectedDevices([SensorTagTemperatureUuids.temperatureService]);
-    peripherals.forEach((peripheral) => log("\t${peripheral.toString()}"));
-  }
-
-  Future<void> _fetchKnownDevice() async {
-    log("Fetch known devices");
-    List<Peripheral> peripherals =
-        await bleManager.knownDevices([peripheral.identifier]);
-    peripherals.forEach((peripheral) => log("\t${peripheral.toString()}"));
+  Future<void> runTestScenario() async {
```
bartoszwilk: This scenario will fail due to unimplemented features of simulator

PR #216, opened 2019-10-11
[iOS][Blemulator] SimulatorPlugin

TomoLV:


ios/Classes/Blemulator/Constants/SimulationChannelName.m
```diff
@@ -0,0 +1,4 @@
+#define BASE @"com.polidea.blemulator"
+
+NSString * const SIMULATION_CHANNEL_NAME_TO_JAVA = (BASE @"/toJava");
```
pawelByszewski: is there any channel to JAVA?

ios/Classes/Blemulator/SimulatorPlugin.m
```diff
@@ -0,0 +1,68 @@
...
+@property (nonatomic) DartValueHandler *dartValueHandler;
+
+@end
+
+@implementation SimulatorPlugin
+
+// MARK: - Public methods
+
++ (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
+    FlutterMethodChannel *dartToPlatformChannel = [FlutterMethodChannel methodChannelWithName:SIMULATION_CHANNEL_NAME_TO_JAVA
```
pawelByszewski: what for is the channel to JAVA?

ios/Classes/Blemulator/Constants/SimulationChannelName.m
TomoLV: It's a Dart channel name that was previously introduced when development was ongoing for Android version, so I'm just sticking to the convention here.

 It was discussed that in the future it might need to be changed to "toPlatform" to be more appropriate for Multiplatform solution (might need to create a separate issue for that task).

ios/Classes/Blemulator/SimulatorPlugin.m
TomoLV: See comment above.

ios/Classes/Blemulator/Constants/SimulationChannelName.m
pawelByszewski: as I understand the value cannot be changed but I suppose the name(`SIMULATION_CHANNEL_NAME_TO_JAVA`) could be, am I right?

TomoLV: Good point, changed to `SIMULATION_CHANNEL_TO_PLATFORM`.

ios/Classes/Blemulator/SimulatorPlugin.h
```diff
@@ -0,0 +1,7 @@
+#import <Flutter/Flutter.h>
+
+@interface SimulatorPlugin : NSObject<FlutterPlugin>
+
+- (instancetype)initWithPlatformToDartChannel:(FlutterMethodChannel *)platformToDartChannel;
```
mikolak: Is it necessary?

ios/Classes/Common/FlutterMethodCallHandler.h
```diff
@@ -0,0 +1,7 @@
+#import <Flutter/Flutter.h>
+
+@protocol FlutterMethodCallHandler
+
+- (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result;
```
mikolak: We don't want FlutterBleLib and BLEmulator to have `Common`, since the latter is to be extracted to a separate repository.

TomoLV: This is a temporary solution until Blemulator is extracted from the repo. My plan was to place all "Common" code here so I can know later which classes I'm going to need to place in both repos.

ios/Classes/Blemulator/SimulatorPlugin.h
TomoLV: It is not 'necessary' but this conformance automatically adds `+ (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar` which is required in the solution.

Other solution would be to just define similar method in the class and just specify `NSObject`  as a base class.

ios/Classes/Common/FlutterMethodCallHandler.h
TomoLV: Created a separate issue for that to address that after currently opened PRs are merged

PR #226, opened 2019-10-11
RSSI in Simulator

pawelByszewski:


lib/blemulator/simulation/simulated_peripheral.dart
```diff
@@ -135,4 +135,6 @@ abstract class SimulatedPeripheral {
     }
     return targetCharacteristic;
   }
+
+  int rssi() => scanInfo.rssi;
```
bartoszwilk: Lets make it return Future<int> as this operation should be asynchronous

pawelByszewski: why should it be asynchronous?

lib/blemulator/internal/simulation_manager_mixins/rssi_mixin.dart
```diff
@@ -0,0 +1,12 @@
+part of internal;
+
+mixin PeripheralRssiMixin on SimulationManagerBaseWithErrorChecks {
+  Map<String, StreamSubscription> _connectionStateSubscriptions = Map();
+
+  Future<int> _readRssiForDevice(String identifier) async {
+    await _errorIfUnknown(identifier);
```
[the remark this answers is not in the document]
pawelByszewski: there is already bug for that to introduce a mechanism for that to whole simulator

lib/blemulator/simulation/simulated_peripheral.dart
bartoszwilk: In standard BT stack, obtaining actual rssi from device is an asynchronous operation because its being evaluated again (with no matter what was the rssi value during scanning process).

PR #227, opened 2019-10-12
monitor transactionId of monitoring characteristics - remove doubled …

pawelByszewski:
…events

### DIFF
--- a/android/src/main/java/com/polidea/flutter_ble_lib/delegate/CharacteristicsDelegate.java
+++ b/android/src/main/java/com/polidea/flutter_ble_lib/delegate/CharacteristicsDelegate.java
@@ -134,14 +134,14 @@ public class CharacteristicsDelegate extends CallDelegate {
 
     private void readCharacteristicForIdentifier(
             int characteristicIdentifier,
-            String transactionId,
+            final String transactionId,
             final MethodChannel.Result result) {
         final SafeMainThreadResolver<Characteristic> safeMainThreadResolver = new SafeMainThreadResolver<>(
                 new OnSuccessCallback<Characteristic>() {
                     @Override
                     public void onSuccess(Characteristic data) {
                         try {
-                            result.success(characteristicsResponseJsonConverter.toJson(createCharacteristicResponse(data)));
+                            result.success(characteristicsResponseJsonConverter.toJson(createCharacteristicResponse(data, transactionId)));
                         } catch (JSONException e) {
                             e.printStackTrace();
                             result.error(null, e.getMessage(), null);
@@ -173,7 +173,7 @@ public class CharacteristicsDelegate extends CallDelegate {
             String deviceIdentifier,
             String serviceUuid,
             String characteristicUuid,
-            String transactionId,
+            final String transactionId,
             final MethodChannel.Result result) {
 
         final SafeMainThreadResolver<Characteristic> safeMainThreadResolver = new SafeMainThreadResolver<>(
@@ -181,7 +181,7 @@ public class CharacteristicsDelegate extends CallDelegate {
                     @Override
                     public void onSuccess(Characteristic data) {
                         try {
-                            result.success(characteristicsResponseJsonConverter.toJson(createCharacteristicResponse(data)));
+                            result.success(characteristicsResponseJsonConverter.toJson(createCharacteristicResponse(data, transactionId)));
                         } catch (JSONException e) {
                             e.printStackTrace();
                             result.error(null, e.getMessage(), null);
@@ -209,13 +209,13 @@ public class CharacteristicsDelegate extends CallDelegate {
                 });
     }
 
-    private void readCharacteristicForService(int serviceIdentifier, String characteristicUuid, String transactionId, final MethodChannel.Result result) {
+    private void readCharacteristicForService(int serviceIdentifier, String characteristicUuid, final String transactionId, final MethodChannel.Result result) {
         final SafeMainThreadResolver<Characteristic> safeMainThreadResolver = new SafeMainThreadResolver<>(
                 new OnSuccessCallback<Characteristic>() {
                     @Override
                     public void onSuccess(Characteristic data) {
                         try {
-                            result.success(characteristicsResponseJsonConverter.toJson(createCharacteristicResponse(data)));
+                            result.success(characteristicsResponseJsonConverter.toJson(createCharacteristicResponse(data, transactionId)));
                         } catch (JSONException e) {
                             e.printStackTrace();
                             result.error(null, e.getMessage(), null);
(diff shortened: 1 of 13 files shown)


lib/characteristic.dart
```diff
@@ -76,4 +77,11 @@ class CharacteristicWithValue extends Characteristic with WithValue {
   ) : super.fromJson(jsonObject, service, manager) {
     value = base64Decode(jsonObject[_CharacteristicMetadata.value]);
   }
+
+  CharacteristicWithValue setTransactionId(String transactionId) {
```
bartoszwilk: What about passing transactionId to characteristicResponse from native side and only filtering it in CharacteristicMixin by obtaining the transactionId from raw json? I think that characteristic entity shouldn't be aware of our "pending transactions" mechanism.

pawelByszewski: to be strict, characteristic entity is not aware of transaction id but there is a DTO that contains characteristic along with value and transactionID. I introduced additional mixin to separate transactionId from the previous model. If I would follow your suggestion I would have to introduced addition DTO anyway.

android/src/main/java/com/polidea/flutter_ble_lib/delegate/CharacteristicsDelegate.java
```diff
@@ -173,15 +173,15 @@ private void readCharacteristicForDevice(
...
+            final String transactionId,
             final MethodChannel.Result result) {
 
         final SafeMainThreadResolver<Characteristic> safeMainThreadResolver = new SafeMainThreadResolver<>(
                 new OnSuccessCallback<Characteristic>() {
                     @Override
                     public void onSuccess(Characteristic data) {
                         try {
-                            result.success(characteristicsResponseJsonConverter.toJson(createCharacteristicResponse(data)));
+                            result.success(characteristicsResponseJsonConverter.toJson(createCharacteristicResponse(data, transactionId)));
```
mikolak: We only need to add transactionId to the one sent through EventChannel.

lib/characteristic.dart
```diff
@@ -68,12 +68,23 @@ mixin WithValue on Characteristic {
   Uint8List value;
 }
 
-class CharacteristicWithValue extends Characteristic with WithValue {
-  CharacteristicWithValue.fromJson(
+mixin WitTransactionId on Characteristic {
```
mikolak: `With`

lib/characteristic.dart
```diff
@@ -68,12 +68,23 @@ mixin WithValue on Characteristic {
...
+  String transactionId;
+}
+
+class CharacteristicWithValueWithTransactionId extends Characteristic with WithValue, WitTransactionId {
```
mikolak: `CharacteristicWithValue` and `CharacteristicWithValueAndTransactionId`? One with transactionId is only need for monitoring, and then the mixin could be on `CharacteristicWithValue`.

mikolak: Actually we might not need mixins at all, since we're declaring classes with them in the same file and they're not holding any logic.

lib/internal/bridge/characteristics_mixin.dart
```diff
@@ -21,9 +21,9 @@ mixin CharacteristicsMixin on FlutterBLE {
           .catchError((errorJson) =>
               Future.error(BleError.fromJson(jsonDecode(errorJson.details))))
           .then((rawJsonValue) =>
-              _parseCharacteristicWithValueResponse(peripheral, rawJsonValue).value);
+              _parseCharacteristicWithValueWithTransactionIdResponse(peripheral, rawJsonValue).value);
 
-  Future<CharacteristicWithValue> readCharacteristicForDevice(
+  Future<CharacteristicWithValueWithTransactionId> readCharacteristicForDevice(
```
mikolak: Doesn't need transactionId.

lib/service.dart
```diff
@@ -32,7 +32,7 @@ class Service extends InternalService {
       _manager.writeCharacteristicForService(peripheral, this,
           characteristicUUID, bytes, withResponse, transactionId);
 
-  Future<CharacteristicWithValue> readCharacteristic(String characteristicUUID,
+  Future<CharacteristicWithValueWithTransactionId> readCharacteristic(String characteristicUUID,
```
mikolak: No need to change type here.

lib/peripheral.dart
```diff
@@ -56,7 +56,7 @@ class Peripheral {
     return _manager.requestMtu(this, mtu, transactionId);
   }
 
-  Future<CharacteristicWithValue> readCharacteristic(
+  Future<CharacteristicWithValueWithTransactionId> readCharacteristic(
```
mikolak: No need to change type here.

lib/internal/managers_for_classes.dart
```diff
@@ -42,7 +42,7 @@ abstract class ManagerForPeripheral {
       bool withResponse,
       String transactionId);
 
-  Stream<CharacteristicWithValue> monitorCharacteristicForDevice(
+  Stream<CharacteristicWithValueWithTransactionId> monitorCharacteristicForDevice(
```
mikolak: Also: transactionId doesn't need to be visible to the user, it's an internal mechanism of the library.

lib/peripheral.dart
```diff
@@ -84,7 +84,7 @@ class Peripheral {
         transactionId,
       );
 
-  Stream<CharacteristicWithValueWithTransactionId> monitorCharacteristic(
+  Stream<CharacteristicWithValueAndTransactionId> monitorCharacteristic(
```
mikolak: `CharacteristicWithValue`

lib/service.dart
```diff
@@ -41,7 +41,7 @@ class Service extends InternalService {
         transactionId,
       );
 
-  Stream<CharacteristicWithValueWithTransactionId> monitorCharacteristic(
+  Stream<CharacteristicWithValueAndTransactionId> monitorCharacteristic(
```
mikolak: CharacteristicWithValue

lib/internal/bridge/characteristics_mixin.dart
```diff
@@ -163,7 +163,7 @@ mixin CharacteristicsMixin on FlutterBLE {
             throw BleError.fromJson(jsonDecode(errorJson.details)));
   }
 
-  Stream<CharacteristicWithValueWithTransactionId> monitorCharacteristicForDevice(
+  Stream<CharacteristicWithValueAndTransactionId> monitorCharacteristicForDevice(
```
mikolak: `CharacteristicWithValue`

lib/internal/bridge/characteristics_mixin.dart
```diff
@@ -194,7 +194,7 @@ mixin CharacteristicsMixin on FlutterBLE {
             throw BleError.fromJson(jsonDecode(errorJson.details)));
   }
 
-  Stream<CharacteristicWithValueWithTransactionId> monitorCharacteristicForService(
+  Stream<CharacteristicWithValueAndTransactionId> monitorCharacteristicForService(
```
mikolak: `CharacteristicWithValue`

lib/characteristic.dart
```diff
@@ -72,19 +72,29 @@ mixin WitTransactionId on Characteristic {
...
     Service service,
     ManagerForCharacteristic manager,
   ) : super.fromJson(jsonObject, service, manager) {
     value = base64Decode(jsonObject[_CharacteristicMetadata.value]);
   }
 
-  CharacteristicWithValueWithTransactionId setTransactionId(String transactionId) {
+}
+
+class CharacteristicWithValueAndTransactionId extends CharacteristicWithValue with WitTransactionId {
```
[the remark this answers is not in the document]
mikolak: BTW I think both mixin and this type should be move to internal package/library.

PR #228, opened 2019-10-12
Jitpack

pawelByszewski:


android/settings.gradle
```diff
@@ -1,4 +1 @@
 rootProject.name = 'flutter_ble_lib'
-
-include ':multiplatformbleadapter'
```
mikolak: This should be removed from `example/android/settings.gradle` to see if it works

PR #229, opened 2019-10-13
an app can connect to a device more than once

pawelByszewski:


bartoszwilk:
Connection state events are duplicated if you connect/disconnect couple of times. Could you check if it is related with our already known bug about duplicating events? From what I observed we get only one callback invocation from native side. The duplication seems to be caused by flutter side.
Screenshot 2019-10-14 at 11 16 03

pawelByszewski:
nope, it is another bug. WIll fix

pawelByszewski:
The bug is connected with using of the sample in the app, it is fixed in PR to develop

PR #233, opened 2019-10-14
add option to stop observing connection state on disconnect

pawelByszewski:


lib/peripheral.dart
```diff
@@ -30,8 +30,30 @@ class Peripheral {
           timeout: timeout);
 
   Stream<PeripheralConnectionState> observeConnectionState(
-          {bool emitCurrentValue = false}) =>
-      _manager.observePeripheralConnectionState(identifier, emitCurrentValue);
+          {bool emitCurrentValue = false,
+          bool unsubscribeOnCancel = false}) {
```
mikolak: `completeOnDisconnect`?

lib/peripheral.dart
```diff
@@ -30,8 +30,30 @@ class Peripheral {
...
+
+        var streamTransformer = StreamTransformer<PeripheralConnectionState, PeripheralConnectionState>.fromHandlers(
```
mikolak: Is it needed? Couldn't it be handled by 
```
return _manager.observePeripheralConnectionState(identifier, emitCurrentValue)
   .takeWhile(
       (state) { 
           if (completeOnDisconnect) 
               return state != PeripheralConnectionState.disconnected 
           else 
               return true; 
        })
```

mikolak: If the answer is no, then please extract this behaviour to InternalBleManager and just forward the argument from here to there.

android/src/main/java/com/polidea/flutter_ble_lib/delegate/DeviceConnectionDelegate.java
```diff
@@ -1,6 +1,7 @@
 package com.polidea.flutter_ble_lib.delegate;
 
 import android.support.annotation.NonNull;
+import android.util.Log;
```
mikolak: Unused import

PR #241, opened 2019-10-15
[Simulator] canceling pending transactions

bartoszwilk:


android/src/main/java/com/polidea/blemulator/bridging/DartMethodCaller.java
```diff
@@ -220,11 +220,12 @@ private BleError objectToBleError(Object o) {
     public void discoverAllServicesAndCharacteristicsForDevice(
             final String deviceIdentifier,
             final String name,
-            String transactionId,
+            final String transactionId,
             final OnSuccessCallback<DeviceContainer> onSuccessCallback,
             final OnErrorCallback onErrorCallback) {
         dartMethodChannel.invokeMethod(DartMethodName.DISCOVER_ALL_SERVICES_AND_CHARACTERISTICS, new HashMap<String, Object>() {{
             put(ArgumentName.IDENTIFIER, deviceIdentifier);
+            put(SimulationArgumentName.TRANSACTION_ID, transactionId);
```
pawelByszewski: why don't you use `ArgumentKey`?

android/src/main/java/com/polidea/blemulator/bridging/DartMethodCaller.java
```diff
@@ -445,4 +458,11 @@ public void notImplemented() {
                     }
                 });
     }
+
+    public void cancelTransaction(final String transactionId) {
+        Log.i(TAG, "cancelTransaction");
+        dartMethodChannel.invokeMethod(DartMethodName.CANCEL_TRANSACTION, new HashMap<String, String>() {{
+            put(SimulationArgumentName.TRANSACTION_ID, transactionId);
```
pawelByszewski: sometime you use `ArgumentKey.TRANSACTION_ID` and sometimes `SimulationArgumentName.TRANSACTION_ID`.  Why?

mikolak: Both should ArgumentName.TRANSACTION_ID.
SimulationArgumentName is for arguments passed from Dart to platform, and ArgumentKey is from FlutterBleLib.

lib/blemulator/internal/bridge/constants.dart
```diff
@@ -68,6 +70,7 @@ abstract class SimulationArgumentName {
 
   static const String characteristics = "characteristics";
   static const String uuid = "uuid";
+  static const String transactionId = "transactionId";
```
mikolak: This should be in ArgumentName, as it is incoming, not outgoing.

lib/blemulator/internal/bridge/platform_to_dart_bridge.dart
```diff
@@ -3,13 +3,57 @@ part of internal;
...
+    }
+  }
+
+  Future<dynamic> _createAndSaveCancelablePlatformCall(
+    Future<dynamic> platformCallResult,
+    String transactionId,
+  ) {
+    CancelableOperation operation =
+        CancelableOperation.fromFuture(platformCallResult);
+    pendingTransactions.putIfAbsent(transactionId, () => operation);
```
mikolak: And what if it's not absent?

lib/blemulator/internal/bridge/platform_to_dart_bridge.dart
```diff
@@ -3,13 +3,57 @@ part of internal;
...
 
   PlatformToDartBridge(this._manager) {
     _platformToDartChannel = new MethodChannel(ChannelName.platformToDart);
     _platformToDartChannel.setMethodCallHandler(_handleCall);
   }
 
-  Future<dynamic> _handleCall(MethodCall call) {
+  Future<dynamic> _handleCall(MethodCall call) async {
+    String transactionId = call.arguments[SimulationArgumentName.transactionId];
+    Future<dynamic> platformCallResultFuture = _dispatchPlatformCall(call);
```
mikolak: You're not checking if the called operation even _is_ cancellable, you always add it the map of pending transactions with null key (since non-cancellables won't have `transactionId`).

lib/blemulator/internal/bridge/platform_to_dart_bridge.dart
bartoszwilk: I added cancelling & removing pending transactions if they already existed.

lib/blemulator/internal/bridge/platform_to_dart_bridge.dart
```diff
@@ -11,14 +11,15 @@ class PlatformToDartBridge {
   }
 
   Future<dynamic> _handleCall(MethodCall call) async {
-    String transactionId = call.arguments[SimulationArgumentName.transactionId];
-    Future<dynamic> platformCallResultFuture = _dispatchPlatformCall(call);
-
-    if (transactionId != null) {
+    if (call.method != DartMethodName.cancelTransaction &&
```
mikolak: What do you think about creating `bool isCallCancellable(MethodCall call)`?

lib/blemulator/internal/bridge/platform_to_dart_bridge.dart
```diff
@@ -11,14 +11,15 @@ class PlatformToDartBridge {
...
+        call.arguments.containsKey(SimulationArgumentName.transactionId)) {
+      String transactionId =
+          call.arguments[SimulationArgumentName.transactionId];
+      _cancelTransactionIfExists(transactionId);
       return _createAndSaveCancelablePlatformCall(
```
mikolak: How about changing this method to just accept MethodCall, and pulling the argument out inside? _dispatchPlatformCall could be hidden from here as well then.

PR #259, opened 2019-10-17
[iOS][Blemulator] Read and write characteristic methods

### LINKED ISSUE #249: Read/write characteristics

TomoLV:
Resolves #249


ios/Classes/Blemulator/Bridging/Converter/DartResultConverter.m
```diff
@@ -37,6 +29,37 @@ + (DeviceContainer *)deviceContainerFromDartResult:(id)result
...
+    NSDictionary *resultDictionary = (NSDictionary *)result;
+    return [self characteristicFromDictionary:resultDictionary
+                                      service:[self serviceFromDictionary:resultDictionary
+                                                               peripheral:[self peripheralFromDictionary:resultDictionary]]];
+}
+
++ (Peripheral *)peripheralFromDictionary:(NSDictionary *)dictionary {
+    return [[Peripheral alloc] initWithIdentifier:[dictionary objectForKey:DART_RESULT_DEVICE_IDENTIFIER]
+                                             name:nil
+                                              mtu:23];
```
mikolak: ?

TomoLV: Constructing a Characteristic instance requires providing a Service object which requires Peripheral object as one of its arguments. (on Android it requires only deviceIdentifier and not the whole Peripheral object).

Because it is being used only for passing the value forward (converting from one json to another with Characteristic object representation in the middle of the process) I had to somehow create a "dummy" peripheral object, since the only information that is written to characteristic's json representation is deviceIdentifier.

mikolak: Disclaimer: this has been discussed in person.

tl;dr mtu should not be set in the peripheral and instead should be a function that reaches over to Dart for answers.

PR #284, opened 2019-10-21
Add README to BLEmulator

### LINKED ISSUE #195: Documentation for Simulator

mikolak:
Resolves #195

### DIFF
--- a/lib/blemulator/README.md
+++ b/lib/blemulator/README.md
@@ -85,7 +85,6 @@ ScanInfo supports following data:
 ```dart
 class ScanInfo {
   int rssi;
-  int mtu;
   bool isConnectable;
   int txPowerLevel;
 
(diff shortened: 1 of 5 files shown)


lib/blemulator/README.md
```diff
@@ -0,0 +1,253 @@
...
+For your convenience you can mark SimulatedService as advertised, but be aware that this doesn't validate the size of advertised services array inside advertisement data.
+
+If you want to have more granular control over what is advertised, simply modify the _scanInfo_ property of the SimulatedPeripheral.
+ The _scanInfo_ is broadcasted automatically while peripheral scan is on each _advertisementInterval_.
+ 
+ScanInfo supports following data:
+```dart
+class ScanInfo {
+  int rssi;
+  int mtu;
```
dariuszseweryn: I think it is a mistake by putting `mtu` in `ScanInfo` — MTU is a property of a connection made to peripheral, not scan's property.

lib/blemulator/README.md
```diff
@@ -0,0 +1,253 @@
...
+      this.serviceUuids,
+      this.localName,
+      this.solicitedServiceUuids,
+      this.overflowUuids});
+}
+```
+You can also provide it in the constructor of the SimulatedPeripheral.
+
+#### Custom characteristic behaviour
+Blemulator does most of the heavy lifting for you adn takes care of the basic stuff, but there's always more complicated logic.
```
dariuszseweryn:
```suggestion
Blemulator does most of the heavy lifting for you and takes care of the basic stuff, but there's always more complicated logic.
```

lib/blemulator/README.md
```diff
@@ -0,0 +1,253 @@
...
+If you need to validate values or a write to one characteristic has to trigger a change in the behaviour of another, you may need to extend SimulatedService or SimulatedCharacteristic classes.
```
dariuszseweryn:
```suggestion
If you need to validate values or write to one characteristic that has to trigger a change in behaviour of a different characteristic, you may need to extend SimulatedService or SimulatedCharacteristic classes.
```

lib/blemulator/README.md
```diff
@@ -0,0 +1,253 @@
...
+    scanInfo.localName = localName;
+
+    getCharacteristicForService(_serviceUuid, _temperatureConfigUuid)
+        .monitor()
+        .listen((value) {
+      int valueAsInt = value.buffer.asByteData().getUint8(0);
+      _readingTemperature = valueAsInt == 1 ? true : false;
+    });
+
+    _emitTemperature();
```
dariuszseweryn: Will this practically launch a new thread that will emit values?

mikolak: In a way. While this is the same thread, it runs like a saga in RN. Since the `_emitTemperature()` function is marked as async and there's no `await` keyword before this call, it won't pause the execution of this method.

There are isolates in Dart that are physical threads, but they cannot share any context with the main thread.

lib/blemulator/README.md
mikolak: That's true, it's now removed completely from there.

PR #286, opened 2019-10-21
config for release 2.0.0-dev.1

pawelByszewski:


pubspec.yaml
```diff
@@ -1,8 +1,8 @@
 name: flutter_ble_lib
-description: A new flutter plugin project.
-version: 0.0.1
-author:
-homepage:
+description: FlutterBle Library is a flutter library that supports ble connections. It uses RxAndroidBle and RxBluetoothKit as native libs.
```
mikolak: >...supports BLE operations. It uses MultiPlatformBleAdapter as a native backend.
?

bartoszwilk: "supports BLE operations" also sounds better for me. For me, it is a good idea to inform user that this library is backed by our, popular in community, libraries instead of some mysterious MultiplatformBleAdapter.

PR #292, opened 2019-10-22
[iOS] Support restoreState

TomoLV:


example/lib/devices_list/devices_bloc.dart
```diff
@@ -48,7 +48,15 @@ class DevicesBloc {
     bleDevices.clear();
     Blemulator().addSimulatedPeripheral(SensorTag());
     Blemulator().simulate();
-    _bleManager.createClient()
+    _bleManager.createClient(
+          restoreStateIdentifier: "example-restore-state-identifier",
+          restoreStateAction: (peripherals) {
+            peripherals.forEach((peripheral) {
+              bleDevices.add(ConnectedBleDevice(peripheral.name, peripheral.identifier, peripheral));
```
mikolak: It should be as DisconnectedBleDevice

lib/internal/bridge/lib_core.dart
```diff
@@ -31,7 +31,7 @@ class FlutterBleLib extends FlutterBLE
           if (jsonString == null)
             return null;
           else {
-            List<Map<String, dynamic>> restoredPeripheralsJson =
+            List<dynamic> restoredPeripheralsJson =
                 jsonDecode(jsonString);
```
mikolak: `List<Map<String, dynamic>> restoredPeripheralsJson =
            (jsonDecode(jsonString) as List<dynamic>).cast();`

TomoLV: Changed as requested

example/lib/devices_list/devices_bloc.dart
TomoLV: Changed to BleDevice.notConnected

PR #311, opened 2019-10-28
Fix duplicated monitoring events

### LINKED ISSUE #258: [FlutterBleLib] Monitored characteristics from `monitorCharacteristicForDevice()` are not filtered by `peripheralId`, which might cause multiple readings when monitoring characteristics with same service and characteristic UUID on different peripherals.

bartoszwilk:
Fixes #258


pawelByszewski:
what about some unit tests?

PR #313, opened 2019-10-29
[iOS] Rename ArgumentValidator to ArgumentChecker

### LINKED ISSUE #181: Rename ArgumentValidator class and its calls

TomoLV:
Resolves #181


ios/Classes/Util/ArgumentChecker.h
```diff
@@ -0,0 +1,11 @@
+@interface ArgumentChecker : NSObject
```
mikolak: It's not checking, is it? Parser, Caster, Handler would all be all right, but the way I see it, this doesn't check for anything.

TomoLV: My understanding was that if was checking if `argument` was a `[NSNull null]` object (inside if statement), hence the name "Checker", but I agree that "Handler" seems better.

Changed the name to the `ArgumentHandler`.

PR #328, opened 2019-11-08
Simplify and clarify public API

mikolak:


lib/internal/internal_ble_manager.dart
```diff
@@ -156,12 +156,12 @@ class InternalBleManager
...
   @override
-  Future<List<Peripheral>> knownDevices(List<String> peripheralIdentifiers) {
+  Future<List<Peripheral>> knownPeripherals(List<String> peripheralIdentifiers) {
     return _bleLib.knownDevices(peripheralIdentifiers ?? []);
   }
 
   @override
-  Future<List<Peripheral>> connectedDevices(List<String> serviceUUIDs) {
+  Future<List<Peripheral>> connectedPeripherals(List<String> serviceUUIDs) {
     return _bleLib.connectedDevices(serviceUUIDs ?? []);
```
TomoLV: Maybe lib needs to address names of the methods as well? (`_bleLib.connectedDevices`) - subject for a separate issue.

mikolak: _bleLib is an internal implementation detail. We might change anytime or decide against it, either way it won't affect the users.

TomoLV: Good point

PR #334, opened 2019-12-02
Lock MBA version on Android

mikolak:


android/build.gradle
```diff
@@ -36,5 +36,5 @@ android {
 
 dependencies {
     implementation 'com.android.support:support-annotations:28.0.0'
-    implementation 'com.github.Polidea:MultiPlatformBleAdapter:master-SNAPSHOT'
+    implementation 'com.github.Polidea:MultiPlatformBleAdapter:dd27a55159fc45956a802520b2cdd1144a34c6ad'
```
pawelByszewski: IMHO it would be better to add a dependency to specific tag instead of commit

mikolak: Done.

PR #340, opened 2019-12-05
upgrade to androidx

hanabi1224:


example/.gitignore
```diff
@@ -64,6 +65,7 @@
 **/ios/ServiceDefinitions.json
 **/ios/Runner/GeneratedPluginRegistrant.*
 **/ios/Flutter/flutter_export_environment.sh
+**/ios/Podfile
```
hanabi1224: In upcoming flutter versions, it requires user to rm ios/Podfile and regenerates on build (with Podfile.lock kept), so I remove the file and gitignore it

hanabi1224:
Looks like ios build is flaky, only succeeded once then keep timing out

TomoLV:
> Looks like ios build is flaky, only succeeded once then keep timing out

@hanabi1224 We have noticed that iOS build is somehow unstable. Sometimes it does stall indeed. Ive restarted the job.

example/.gitignore
dariuszseweryn: Could you point to a relevant discussion on flutter site so I could read more? I cannot google it

hanabi1224:
@TomoLV Thanks!  I feel cirrus ci (which flutter official plugin repo uses) is more stable for E2E build.  I can send a PR to enable cirrus CI if you like

TomoLV:
> @TomoLV Thanks! I feel cirrus ci (which flutter official plugin repo uses) is more stable for E2E build. I can send a PR to enable cirrus CI if you like

Thanks for a suggestion. I'm not sure if it's necessary right now - we need to have a deeper look into this unstable iOS build on Travis CI and decide what to do with this issue.

example/.gitignore
hanabi1224: @dariuszseweryn Ah... I don't think it's documented anywhere but if you switch flutter channel to beta or dev, then run cd example && flutter build ios --no-codesign .  it will warn that you are using an outdated Podfile, run rm ios/Podfile and build again

hanabi1224:
@pawelByszewski Thanks for the feedback!  I totally agree with the philosophy to keep the PR minimal with a single purpose, however, most of the changes only apply to the example project, how about separating it into 2? One with changes made to the library, and one with all changes to the example?

PR #350, opened 2019-12-12
Added a UTM link to README

bntnam:
Hi,
According to the Communication team's tasks, I added a UTM link into README which linked to the contact form in Polidea website. Thus, we could track people how they get to know us.
Thanks from Nam Bui.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,8 @@ Monitoring or reading a characteristic from _Peripheral_/_Service_ level
 ## Maintained by
 This library is maintained by [Polidea](http://www.polidea.com)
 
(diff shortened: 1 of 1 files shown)


mikolak:
@bntnam should tracking links also be added to other links? The line above and the line below also link to Polidea's site.

bntnam:
> @bntnam should tracking links also be added to other links? The line above and the line below also link to Polidea's site.
I have no idea about the above and below links. The one I added which included something like a campaign ID. So when users submit the form, based on that, we'll know where it comes from. And I was assigned to implement it from Ula. 😄

PR #353, opened 2019-12-12
fix transaction id generator

### LINKED ISSUE #352: TransactionIdGenerator.getNextId - Missing starting value for integer.
https://github.com/Polidea/FlutterBleLib/blob/3507cbc5952b358908d42fdb13a70087a708e6b2/lib/internal/util/transaction_id_generator.dart#L2

This should have a default value. Otherwise `NoSuchMethodError: The method '+' was called on null.`
```dart
static int _id = 0;
```

pawelByszewski:
fixes #352

### DIFF
--- a/test/internal/util/transcation_id_generator_test.dart
+++ b/test/internal/util/transcation_id_generator_test.dart
@@ -1,4 +1,3 @@
-
 import 'package:flutter_test/flutter_test.dart';
 import 'package:flutter_ble_lib/internal/util/transaction_id_generator.dart';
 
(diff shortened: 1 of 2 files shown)


test/internal/util/transcation_id_generator_test.dart
```diff
@@ -0,0 +1,9 @@
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter_ble_lib/internal/util/transaction_id_generator.dart';
+
+void main() {
+  test("should be able to generate an id", () {
+    expect(TransactionIdGenerator.getNextId(), isNotNull);
+  });
+}
```
dariuszseweryn: Maybe also check if next call will not return the previous result?

bartoszwilk: I added a test that checks if next invocation returns incremented id so it also covers your case.

pawelByszewski: The generator needs more test but this time I added test only for the fixed scenario with full determination.

PR #371, opened 2020-01-08
Remove setting MTU to 23 in connect by default

### LINKED ISSUE #361: Can't get the mtu value from the connection.
I can't find in the source code to get the mtu value from the connection.
The requestMtu() will get the invalid PDU error.
Any idea?

@mikolak

_Originally posted by @NOA-Charles in https://github.com/Polidea/FlutterBleLib/issues/315#issuecomment-567773838_

mikolak:
Fixes #361

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -113,9 +113,13 @@ All of the following methods belong to Peripheral instance.
           Duration timeout});
 ```
 Attempts to connect to the peripheral.
+
 `autoConnect` waits for device to be discoverable before attempting connection; Android-specific. [See more](https://github.com/Polidea/RxAndroidBle#auto-connect) 
-`requestMtu` defaults to 23.
+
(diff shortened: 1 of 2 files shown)


REFERENCE.md
```diff
@@ -113,9 +113,13 @@ All of the following methods belong to Peripheral instance.
           Duration timeout});
 ```
 Attempts to connect to the peripheral.
+
 `autoConnect` waits for device to be discoverable before attempting connection; Android-specific. [See more](https://github.com/Polidea/RxAndroidBle#auto-connect) 
-`requestMtu` defaults to 23.
+
+`requestMtu` defaults to 0, which means the library won't set it. Valid values according to BLE specification are between 23 and 512. **IMPORTANT** can be requested only once: if you pass this argument, then every call to`requestMtu()` will fail.
```
bartoszwilk:
```suggestion
`requestMtu` defaults to 0, which means the library won't set it. Valid values according to BLE specification are between 23 and 512. **IMPORTANT** can be requested only once: if you pass this argument, then every call to `requestMtu()` will fail. 
```

PR #375, opened 2020-01-09
Fix incorrect Manufacturer's Data

### LINKED ISSUE #366: Manufacturer Data is not correct
I'm debugging with BlueSee and cross referencing when I port the implementation to a Flutter app using this library. I need a way to identify devices from certain manufacturers (not the specific device but a signature of the type of device so I know which model it is, etc). So I was going to use manufacturer data, but it's not working quite right.

- First, it's an `Int8List` and I would have expected a `Uint8List` so we can get the raw bytes and parse them.
- More concerning is that I just tested with 4 different devices, and for each, I'm getting the same Manufacturer Data (see below) across 4 difference devices when using this library but BlueSee is showing me very different data... so I think there is a bug here.

What I'm getting on all of them is this (even though BlueSee shows different info):
`[-103, -87, -18, 125, -89, 45, -70, -73, -85, 13, -85, 90]`

I'm testing on a physical device (`iPhone 11 Pro Max` running `iOS 13.3`).

mikolak:
Fixes #366

### DIFF
--- a/lib/internal/bridge/scanning_mixin.dart
+++ b/lib/internal/bridge/scanning_mixin.dart
@@ -14,6 +14,10 @@ mixin ScanningMixin on FlutterBLE {
     List<String> uuids,
     bool allowDuplicates,
   ) async* {
+    if (_scanEvents == null) {
+      _prepareScanEventsStream();
+    }
+
     _methodChannel.invokeMethod(
       MethodName.startDeviceScan,
       <String, dynamic>{
@@ -23,10 +27,6 @@ mixin ScanningMixin on FlutterBLE {
         ArgumentName.allowDuplicates: allowDuplicates,
       },
     );
-
-    if (_scanEvents == null) {
-      _prepareScanEventsStream();
-    }
 
     yield* _scanEvents.handleError(
       (errorJson) {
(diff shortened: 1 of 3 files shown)


lib/internal/bridge/scanning_mixin.dart
```diff
@@ -24,10 +28,6 @@ mixin ScanningMixin on FlutterBLE {
       },
     );
 
-    if (_scanEvents == null) {
```
dariuszseweryn: is this related to the issue?

mikolak: Reversed.

mikolak:
@pawelByszewski @dariuszseweryn I've added tests.

test/scan_result.dart
```diff
@@ -0,0 +1,106 @@
...
+            _createJsonScanResult(serviceData: serviceData);
+
+        ScanResult scanResult =
+            ScanResult.fromJson(jsonDecode(serializedScanResult), null);
+        expect(scanResult.advertisementData.serviceData, equals(serviceData));
+      });
+    }
+
+    testServiceDataDeserialization({
+      "1": Uint8List.fromList([0, 1, 2]),
```
pawelByszewski: for me, it is not readable at all, I dunno what `1` or `2` means

PR #378, opened 2020-01-14
Delay start of characteristic monitoring until sink is created

### LINKED ISSUE #358: about indicate
i use this library to receive the packet by other bluetooth device indicate ,but sometimes it will lost some packet;

mikolak:
Fixes #358

### DIFF
--- a/lib/internal/bridge/characteristics_mixin.dart
+++ b/lib/internal/bridge/characteristics_mixin.dart
@@ -155,8 +155,8 @@ mixin CharacteristicsMixin on FlutterBLE {
           },
         );
 
-    bool Function(CharacteristicWithValueAndTransactionId) characteristicFilter =
-        (characteristic) =>
+    bool Function(CharacteristicWithValueAndTransactionId)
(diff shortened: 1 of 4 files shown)


lib/internal/bridge/characteristics_mixin.dart
```diff
@@ -178,27 +175,29 @@ mixin CharacteristicsMixin on FlutterBLE {
...
+          MethodName.monitorCharacteristicForDevice,
+          <String, dynamic>{
+            ArgumentName.deviceIdentifier: peripheral.identifier,
+            ArgumentName.serviceUuid: serviceUuid,
+            ArgumentName.characteristicUuid: characteristicUUID,
+            ArgumentName.transactionId: transactionId,
+          },
+        );
+
+    bool Function(CharacteristicWithValueAndTransactionId) filter =
```
pawelByszewski: what about some more meaningful name? filter what?

bartoszwilk:
What about tests that check if transaction is cancelled after last subscriber unsubscribes on particular transactionId?

PR #379, opened 2020-01-17
Release 2.1.0

mikolak:
I'll later merge master into develop.

## 2.1.0

* **BREAKING** ScanResult.AdvertisementData.ManufacturerData has changed type from Int8List to Uint8List
* Fix issue with invalid value of manufacturer data in scan results
* Fix issue with initialising plugin from a service without an active activity
* Update README with information about permissions
* Fix issue with default value of MTU when connecting to a peripheral on Android
* Fix issue where first few notifications right after start of monitoring a characteristic could get lost

### DIFF
--- a/android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
+++ b/android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
@@ -61,7 +61,7 @@ public class FlutterBleLibPlugin implements MethodCallHandler {
         final EventChannel connectionStateChannel = new EventChannel(registrar.messenger(), ChannelName.CONNECTION_STATE_CHANGE_EVENTS);
         final EventChannel characteristicMonitorChannel = new EventChannel(registrar.messenger(), ChannelName.MONITOR_CHARACTERISTIC);
 
-        final FlutterBleLibPlugin plugin = new FlutterBleLibPlugin(registrar.activity().getApplicationContext());
+        final FlutterBleLibPlugin plugin = new FlutterBleLibPlugin(registrar.context());
 
         channel.setMethodCallHandler(plugin);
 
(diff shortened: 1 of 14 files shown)


android/src/main/java/com/polidea/flutter_ble_lib/FlutterBleLibPlugin.java
```diff
@@ -61,7 +61,7 @@ public static void registerWith(Registrar registrar) {
         final EventChannel connectionStateChannel = new EventChannel(registrar.messenger(), ChannelName.CONNECTION_STATE_CHANGE_EVENTS);
         final EventChannel characteristicMonitorChannel = new EventChannel(registrar.messenger(), ChannelName.MONITOR_CHARACTERISTIC);
 
-        final FlutterBleLibPlugin plugin = new FlutterBleLibPlugin(registrar.activity().getApplicationContext());
+        final FlutterBleLibPlugin plugin = new FlutterBleLibPlugin(registrar.context());
```
dariuszseweryn: Why not `getApplicationContext()`?

mikolak: This is, in fact, application context.
https://api.flutter.dev/javadoc/io/flutter/plugin/common/PluginRegistry.Registrar.html#context--

dariuszseweryn: We are not yet up to the new v2 API I suppose?

bartoszwilk: Maybe it's a good idea to migrate android plugin to v2 as it's a minor release? What do you guys think?

dariuszseweryn: We can do that in `2.2.0` no worries ;)

PR #404, opened 2020-02-07
Handle location permission in example on Android

lukerayman:
This PR adds location permission check before scanning on Android. To achieve this, I have added [permission handler](https://pub.dev/packages/permission_handler) plugin. But there are more changes here.

**BleDevice**
I found this class having some unused subclasses, methods and fields. I removed all the unused stuff, so it's now much simpler and easier to understand. For example, `_isConnected` was always `false` and never used, `ConnectedBleDevice` was exactly the same as `BleDevice` and `DisconnectedBleDevice` class had only different `toString()` method and never used `_devicesInConnectingProcess` list. I have also changed constructor. I found it simpler to create `BleDevice` from `ScanResult` instead of 3 different args that all come from one place.

**Visual Studio Code and Windows**
Some of the changes are just formatting done by Visual Studio Code by itself. I have also added `launch.json` file to run the example in VSC without opening native projects. There are some little changes in Android project that I had to make in order to compile it in Windows. You will find newer version of Gradle, and enabled `useAndroidX` and `enableJetifier` flags (they are like this by default in new Flutter projects). 

**Permission Handler**
I implemented location permission check in Devices BLoC. It's one async method that is running just before scanning: `_checkPermissions()`. If the user will not grand location permission on app start, he will have a second chance by using pull to refresh.


example/android/build.gradle
```diff
@@ -5,15 +5,14 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.1'
+        classpath 'com.android.tools.build:gradle:3.5.0'
```
mikolak: Should we bump it for the library as well?

example/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-all.zip
```
mikolak: Should we bump it for the library as well?

example/lib/device_details/device_details_bloc.dart
```diff
@@ -38,22 +38,23 @@ class DeviceDetailsBloc {
...
 
     _logsController = PublishSubject<List<DebugLog>>();
 
     log = (text) {
       var now = DateTime.now();
-      _logs.insert(0, DebugLog('${now.hour}:${now.minute}:${now.second}.${now.millisecond}', text));
+      _logs.insert(
+          0,
+          DebugLog('${now.hour}:${now.minute}:${now.second}.${now.millisecond}',
+              text));
```
mikolak:
```suggestion
              text,));
```

mikolak: And run autoformatter after the addition of the comma, should be nicer.

example/lib/device_details/device_details_bloc.dart
```diff
@@ -38,22 +38,23 @@ class DeviceDetailsBloc {
     _deviceController = BehaviorSubject<BleDevice>.seeded(device);
 
     _connectionStateController =
-        BehaviorSubject<PeripheralConnectionState>.seeded(device.isConnected
-            ? PeripheralConnectionState.connected
-            : PeripheralConnectionState.disconnected);
+        BehaviorSubject<PeripheralConnectionState>.seeded(
+            PeripheralConnectionState.disconnected);
```
mikolak: Is it no longer possible that the peripheral is connected already when entering this screen?

example/lib/device_details/device_details_bloc.dart
```diff
@@ -232,7 +233,8 @@ class DeviceDetailsBloc {
   }
 
   void dispose() async {
-    _deviceController.value?.abandon();
+    await connectionSubscription.cancel();
```
mikolak: I can't find any place where this field would be set in this file?

test/flutter_ble_lib_test.dart
```diff
@@ -1,6 +1,5 @@
 import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
-import 'package:flutter_ble_lib/flutter_ble_lib.dart';
```
mikolak: Whole file can be removed actually.

example/android/build.gradle
lukerayman: I think so because it's default for the current stable version of Flutter.

example/android/gradle/wrapper/gradle-wrapper.properties
lukerayman: Yes.

lukerayman: I mean, it's not mandatory because everything works. But it will be good be aligned with Flutter.

example/lib/devices_list/devices_bloc.dart
```diff
@@ -44,34 +49,51 @@ class DevicesBloc {
...
+
+  Future<void> _checkPermissions() async {
+    if (Platform.isAndroid) {
+      var permissionStatus = await PermissionHandler()
+          .requestPermissions([PermissionGroup.location]);
+
+      _locationPermissionStatus = permissionStatus[PermissionGroup.location];
+
+      if (_locationPermissionStatus != PermissionStatus.granted) {
+        return Future.error("Location permission not granted");
```
mikolak: Might be a good idea to wrap the string into some kind of exception.

example/lib/model/ble_device.dart
```diff
@@ -1,89 +1,48 @@
...
+  int get hashCode => id.hashCode;
 
   @override
   bool operator ==(other) =>
-      other is BleDevice
-          && this.name != null
-          && other.name != null
-          && compareAsciiLowerCase(this.name, other.name) == 0
-          && this.id == other.id;
+      other is BleDevice &&
```
mikolak: Why the change? I think it was quicker to read with connecting operators on the beginning of the line.

example/android/build.gradle
mikolak: Can you take care of that in this PR?

example/android/gradle/wrapper/gradle-wrapper.properties
mikolak: Can you take care of that in this PR?

example/lib/device_details/device_details_bloc.dart
lukerayman: I found that `device.isConnected` flag was always `false` and was never changing to `true`, so I removed it completely. My guess is that it was just copy-pasted from another project. We can consider handling already connected devices but it sounds like a new task to me.

example/lib/device_details/device_details_bloc.dart
lukerayman: You're right! One more thing to remove. :D I have added `cancel()` because VSC shows a warning when streams are not cancelled.

example/lib/model/ble_device.dart
lukerayman: I didn't change it manually. Auto-format on save. 😄

PR #406, opened 2020-02-13
fix `flutter analyze` issues

pawelByszewski:


mikolak:
Actually, what does this fix?

pawelByszewski:
@mikolak, as you can see the variable `subscription1`, was removed, it was unused anyway.

PR #407, opened 2020-02-14
Fix incorrectly implemented method on Android

mikolak:
@TomoLV will provide fix for the same issue on iOS

## Issue:
Retrieving characteristic for service UUID (not ID) was done by retrieving all services, matching service with requested UUID and then calling `getCharacteristicsForService(id)`. 
MBA has a method `getCharacteristicsForDevice(peripheralId, serviceUuid)` which was unused, thus creating more space for errors and risking inconsistency with BLEmulator or react-native-ble-plx.

### DIFF
--- a/android/src/main/java/com/polidea/flutter_ble_lib/MultiCharacteristicsResponse.java
+++ b/android/src/main/java/com/polidea/flutter_ble_lib/MultiCharacteristicsResponse.java
@@ -4,21 +4,34 @@ import com.polidea.multiplatformbleadapter.Characteristic;
 import com.polidea.multiplatformbleadapter.Service;
 
 import java.util.List;
+import java.util.UUID;
 
 public class MultiCharacteristicsResponse {
     private final List<Characteristic> characteristics;
-    private Service service;
+    private int serviceId;
+    private UUID serviceUuid;
 
     public MultiCharacteristicsResponse(List<Characteristic> characteristics, Service service) {
         this.characteristics = characteristics;
-        this.service = service;
+        this.serviceId = service.getId();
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/polidea/flutter_ble_lib/MultiCharacteristicsResponse.java
```diff
@@ -4,21 +4,34 @@
...
-    private Service service;
+    private int serviceId;
+    private UUID serviceUuid;
 
     public MultiCharacteristicsResponse(List<Characteristic> characteristics, Service service) {
         this.characteristics = characteristics;
-        this.service = service;
+        this.serviceId = service.getId();
+        this.serviceUuid = service.getUuid();
+    }
```
dariuszseweryn: How about:
```
    public MultiCharacteristicsResponse(List<Characteristic> characteristics, @Nullable Service service) {
        this.characteristics = characteristics;
        this.serviceId = service != null ? service.getId() : -1;
        this.serviceUuid = service != null ? service.getUuid() : null;
    }
```

dariuszseweryn:
@mikolak Could you add more info about what was the issue in the PR description?

android/src/main/java/com/polidea/flutter_ble_lib/MultiCharacteristicsResponse.java
mikolak: Service in this constructor should not be Nullable. The new constructor is because I'm retrieving service's UUID and ID from field in Characteristic, which does not have a reference to Service object.

PR #409, opened 2020-02-14
Installation guide

lukerayman:
I have added installation guide for both Android and iOS. It is quite detailed because I assumed that some Flutter developers may not have much experience in native development for both platform.

Preview: https://github.com/Polidea/FlutterBleLib/tree/docs/installation

Some of the changes come from auto-formatter in Visual Studio Code.


README.md
```diff
@@ -6,46 +6,108 @@
...
 
 ## BLE Simulator
-This library supports [BLEmulator](https://github.com/Polidea/blemulator_flutter), the BLE simulator. 
+
+This library supports [BLEmulator](https://github.com/Polidea/blemulator_flutter), the BLE simulator.
 The simulation allows one to develop without physical smartphone or BLE peripheral and use one's production BLE–related code in automated testing.
 
+## Installation
+
+To use this plugin, add `flutter_ble_lib` as a [dependency in your pubspec.yaml file](https://pub.dev/packages/flutter_ble_lib/versions/2.1.0).
```
mikolak: This is provided out of the box on pub.dev - https://pub.dev/packages/flutter_ble_lib#-installing-tab-

mikolak: Ah, I see it has been only moved. It can be safely removed.

README.md
```diff
@@ -6,46 +6,108 @@
...
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    package="com.example.your_app">
+
+  <uses-permission android:name="android.permission.BLUETOOTH"/>
+  <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
+
+  ...
+</manifest>
+```
```
mikolak: Permissions are mentioned in Usage section and it bloats the README. If you think it's necessary, perhaps it should be in a separate file, only linked here?

README.md
```diff
@@ -6,46 +6,108 @@
...
+
+### iOS
+
+In Xcode, set iOS Deployment Target (`IPHONEOS_DEPLOYMENT_TARGET`) to 10.0.
```
mikolak: This can be lowered now I think, lemme check.

README.md
lukerayman: Yeah, and I found that every lib I checked on pub.dev has this info in exactly this form.

README.md
lukerayman: Oh, yes, there is a notice about `ACCESS_FINE_LOCATION`. Then I would rather remove it from Usage section and leave it more detailed in Installation. 

For me, from a user perspective, it's more convinient to have the whole instruction inline in README instead of linked to separate file. I don't remember if I ever saw such solution on any GitHub page. As a user I always look for installation instructions on top or README and I get very frustrated when I have to search anywhere else.

README.md
lukerayman: It's a link to this instruction. I for leaving it as it is for users who will find read this README on GitHub.

README.md
```diff
@@ -6,46 +6,106 @@
...
+This library supports [BLEmulator](https://github.com/Polidea/blemulator_flutter), the BLE simulator.
 The simulation allows one to develop without physical smartphone or BLE peripheral and use one's production BLE–related code in automated testing.
 
+## Installation
+
+To use this plugin, add `flutter_ble_lib` as a [dependency in your pubspec.yaml file](https://pub.dev/packages/flutter_ble_lib/versions/2.1.0).
+
+### Android
+
+Set `minSDKVersion` in `[project]/android/app/build.gradle` file to 18.
```
dariuszseweryn: It is optional but below API 18 this will not work at all

README.md
dariuszseweryn: I can create a new patch version `RxAndroidBle 1.7.1` that would have `ACCESS_FINE_LOCATION` and `android:label` removed – Manifest Merger should do the rest. This should make it easier to integrate. WDYT?

README.md
lukerayman: If it won't work at all, then I don't see it optional. 😄

README.md
lukerayman: The less you have to change in native project the better. I think it's a good idea.

mikolak: ACCESS_FINE_LOCATION would still be required for scanning to work, and that's the only way to obtain peripherals for now.

dariuszseweryn: But there would be only a need to have this runtime permission. No need to add it in the manifest because a dependency would have it.

mikolak: Ah, nice one. Yeah, seems to be a good idea!

dariuszseweryn: Technically I have released `1.7.1` which could be integrated in MBA and here. Should help with both manifest entries

lukerayman:
I removed all the unnecessary installation steps. For Android there is just info about changing min SDK version. I checked in 3 different projects and it works.

dariuszseweryn:
If removing `minSdk` is a must for Flutter lib then we could use [this approach](https://stackoverflow.com/questions/27335889/how-do-i-ignore-minsdkversion-of-library-in-android-studio) but we should then add some graceful handling of such situation.

lukerayman:
Why to remove `minSDK`?

dariuszseweryn:
Dunno, to make the integration easier ;P
To be honest I do think that the user would need to specify the `minSdk` at all. It should be merged properly `rxandroidble` => `mba` => `flutter_ble_lib` => user's app. Doesn't it?

lukerayman:
Do you mean removing `minSdkVersion` completely? I just checked it and it looks like the default value is `1`, so still the same error. 😄 I think we can't get rid of it until Flutter changes its Android project template.

dariuszseweryn:
`minSdk` should be 18 because of using MBA which uses RxAndroidBle which states `minSdk 18`. I will come to you

README.md
```diff
@@ -6,46 +6,80 @@
...
+
+```gradle
+defaultConfig {
+  ...
+  // Set min SDK version to 18
+  minSdkVersion 18
+  targetSdkVersion 28
```
mikolak: targetSdkVersion should be trimmed as not related, and then the comment will be needless.

README.md
mikolak:
```suggestion
Support for Bluetooth Low Energy has been added in API 18, hence the library requires `minSDKVersion` to be set to 18. If BLE is not core to your application, [you can override it](https://stackoverflow.com/questions/27335889/how-do-i-ignore-minsdkversion-of-library-in-android-studio) and handle support detection in your code: 

`[project]/android/app.build.gradle`:
```
followed by the snippet

mikolak: Need to check if the link is working in MD.

README.md
lukerayman: I added it to give some context. But I can removed if it's not necessary.

mikolak: minSdkVersion is a required property in build.gradle and I guess everyone will just do a text search for it in the file. :)

README.md
```diff
@@ -6,46 +6,82 @@
...
+defaultConfig {
+  ...
+  minSdkVersion 18
+  ...
+}
+```
+
+Support for Bluetooth Low Energy has been added in API 18, hence the library requires `minSDKVersion` to be set to 18. If BLE is not core to your application, [you can override it](https://stackoverflow.com/questions/27335889/how-do-i-ignore-minsdkversion-of-library-in-android-studio) and handle support detection in your code.
+
+**Notice:** You don't need to add any additional permissions related to BLE because they are already declared in our manifest. However, you still need to request location permission at runtime, which you can implement with [permission_handler](https://pub.dev/packages/permission_handler) package. See [our example](https://github.com/Polidea/FlutterBleLib/tree/master/example) for reference.
```
mikolak:
```suggestion
**Notice:** You don't need to add any permissions related to BLE to the AndroidManifest.xml, because they are already declared in the library's native module. However, you still need to request `ACCESS_FINE_LOCATION` permission at runtime to be able to scan for peripheral. See [example's code](https://github.com/Polidea/FlutterBleLib/blob/develop/example/lib/devices_list/devices_bloc.dart#L80) and [example's pubspec](https://github.com/Polidea/FlutterBleLib/blob/develop/example/pubspec.yaml#L20).
```

README.md
```diff
@@ -6,46 +6,82 @@
...
+defaultConfig {
+  ...
+  minSdkVersion 18
+  ...
+}
+```
+
+Support for Bluetooth Low Energy has been added in API 18, hence the library requires `minSDKVersion` to be set to 18. If BLE is not core to your application, [you can override it](https://stackoverflow.com/questions/27335889/how-do-i-ignore-minsdkversion-of-library-in-android-studio) and handle support detection in your code.
+
+**Notice:** You don't need to add any permissions related to BLE to the `AndroidManifest.xml`, because they are already declared in the library's native module. However, you still need to request `ACCESS_FINE_LOCATION` permission at runtime to be able to scan for peripheral. See [example's code](https://github.com/Polidea/FlutterBleLib/blob/develop/example/lib/devices_list/devices_bloc.dart#L80) and [example's pubspec](https://github.com/Polidea/FlutterBleLib/blob/develop/example/pubspec.yaml#L20).
```
dariuszseweryn: We mustn't link to a specific LOC on `master` as these may change in the future. Link against a specific commit/file/LOC.

lukerayman: Hmm... I'm not sure. If there will be a newer version of `example`, then it feels wrong to me to link to outdated code. I think that README should be update too. However, I understand that it will be hard to constantly maintain it. I don't know.

mikolak: I agree with Darek. If we introduce unrelated change that doesn't outdate the example or this instruction, but moves the line inside the file, then the link will be suddenly outdated and it would be hard to spot.

mikolak:
```suggestion
**Notice:** You don't need to add any permissions related to BLE to the `AndroidManifest.xml`, because they are already declared in the library's native module. However, you still need to request `ACCESS_FINE_LOCATION` permission at runtime to be able to scan for peripheral. See [example's code](https://github.com/Polidea/FlutterBleLib/blob/202c6fe48300be207f80567ca4bee5b6fbc83eb5/example/lib/devices_list/devices_bloc.dart#L80) and [example's pubspec](https://github.com/Polidea/FlutterBleLib/blob/202c6fe48300be207f80567ca4bee5b6fbc83eb5/example/pubspec.yaml#L20).
```

PR #411, opened 2020-02-14
Migrate from Support Library to AndroidX

### LINKED ISSUE #396: Support for AndroidX.

mikolak:
Resolves #396

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.0'
+        classpath 'com.android.tools.build:gradle:3.5.3'
     }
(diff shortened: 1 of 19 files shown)


android/build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.0'
+        classpath 'com.android.tools.build:gradle:3.5.3'
```
lukerayman: It probably doesn't matter but I found that Flutter is stil using `3.5.0` be default.

mikolak: Did it result in any warning?

lukerayman: No, I think it's good.

PR #414, opened 2020-02-17
Example: handle Bluetooth permission on iOS

lukerayman:
When launching the app on iOS for the first time, the system asks for Bluetooth permission. However, the example app ignores it and tries to scan anyway, so it won't work.

This PR fixes this problem by observing Bluetooth state and waiting for `POWERED_ON` value. 

There is also an updated version of Podfile. I got warning when building the app that it needs to be updated for the new version of Flutter:

```
Warning: Podfile is out of date
  This can cause a mismatched version of Flutter to be embedded in your app, which may result in App Store submission rejection or crashes.
  If you have local Podfile edits you would like to keep, see https://github.com/flutter/flutter/issues/24641 for instructions.
To regenerate the Podfile, run:
  rm ios/Podfile
```

 So I removed it and the new auto-generated version is here.

I have also updated .gitignore by adding `Flutter.podspec` to ignore.

### DIFF
--- a/example/lib/devices_list/devices_bloc.dart
+++ b/example/lib/devices_list/devices_bloc.dart
@@ -97,11 +97,15 @@ class DevicesBloc {
 
   Future<void> _waitForBluetoothPoweredOn() async {
     Completer completer = Completer();
-    _bleManager
+    StreamSubscription<BluetoothState> subscription;
+    subscription = _bleManager
         .observeBluetoothState(emitCurrentValue: true)
-        .listen((bluetoothState) {
-      if (bluetoothState == BluetoothState.POWERED_ON && !completer.isCompleted)
+        .listen((bluetoothState) async {
+      if (bluetoothState == BluetoothState.POWERED_ON &&
+          !completer.isCompleted) {
+        await subscription.cancel();
         completer.complete();
+      }
     });
     return completer.future;
   }
--- a/example/lib/devices_list/devices_bloc.dart
+++ b/example/lib/devices_list/devices_bloc.dart
@@ -77,6 +77,11 @@ class DevicesBloc {
   }
 
   Future<void> _checkPermissions() async {
+    _bleManager
+        .observeBluetoothState(emitCurrentValue: true)
+        .listen((bluetoothState) {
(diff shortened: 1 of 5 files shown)


example/lib/devices_list/devices_bloc.dart
```diff
@@ -86,9 +86,30 @@ class DevicesBloc {
...
+    } else if (Platform.isIOS) {
+      /// On iOS the user will be asked to allow Bluetooth access by
+      /// the system. Once it is granted, the Bluetooth state will change from
+      /// UNKNOWN to POWERED_ON. The following function will block the thread
+      /// until it happens.
+      await _waitForBluetoothPoweredOn();
     }
   }
 
+  Future<void> _waitForBluetoothPoweredOn() async {
```
dariuszseweryn: Shouldn't it be in Android related code as well? The scan would not start if bluetooth is off

lukerayman: Implementing this I was having in mind only handling Bluetooth permission on iOS. But you're right it can be used on Android as well, for case when Bluetooth is really turned off and start scanning only after switching it on. I just tested it and it works just like described.

example/lib/devices_list/devices_bloc.dart
```diff
@@ -86,13 +86,13 @@ class DevicesBloc {
...
-      /// UNKNOWN to POWERED_ON. The following function will block the thread
-      /// until it happens.
-      await _waitForBluetoothPoweredOn();
     }
+
+    /// On iOS the user will be asked to allow Bluetooth access by
+    /// the system. Once it is granted, the Bluetooth state will change from
+    /// UNKNOWN to POWERED_ON. The following function will block the thread
+    /// until it happens.
+    await _waitForBluetoothPoweredOn();
```
dariuszseweryn: I also think that it is not directly related to permissions. Permission is OK, the state of adapter is unknown. WDYT?

lukerayman: I moved `_waitForBluetoothPoweredOn()` to the `init()` chain, so it is done as a separate step after checking permissions. I think it looks more clean now. Is it OK for you?

PR #423, opened 2020-02-20
Move example's dependency from plugin to example

### LINKED ISSUE #422: BLE Plugin does not need to have the ability to apply for permissions
"permission_handler" is used in the plugin, I don't think it is good, because "permission_handler" will lead to the need to add extra Purchase String in Info.plist.

Would you consider giving up the ability to gain permissions in this plugin?

mikolak:
Resolves #422

### DIFF
--- a/example/pubspec.yaml
+++ b/example/pubspec.yaml
@@ -17,6 +17,8 @@ dependencies:
   rxdart: ^0.22.0
   fimber: ^0.1.10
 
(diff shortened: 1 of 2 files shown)


lukerayman:
Yeah, it was my mistake...  😅 https://github.com/Polidea/FlutterBleLib/pull/404/files#diff-fb2d52cd953afedcafc9fe7629c6f132

PR #432, opened 2020-02-26
hide private API

### LINKED ISSUE #427: Hide implementation files
https://dart.dev/tools/pub/package-layout#implementation-files

pawelByszewski:
Resolves #427


dariuszseweryn:
Shouldn't all of those file disappear from other paths as well?

pawelByszewski:
> Shouldn't all of those file disappear from other paths as well?

could you specify 'others'

PR #433, opened 2020-02-26
Improvement/docs

pawelByszewski:


lib/scan_result.dart
```diff
@@ -15,14 +15,26 @@ abstract class _ScanResultMetadata {
   static const String overflowServiceUuids = "overflowServiceUUIDs";
 }
 
+/// Type of items returned by [BleManager] during scanning.
```
mikolak:
```suggestion
/// A scan result emitted by the scanning operation, containing [Peripheral] and [AdvertisementData]
```

lib/scan_result.dart
```diff
@@ -15,14 +15,26 @@ abstract class _ScanResultMetadata {
...
+  bool isConnectable;
+
+  /// A list of UUIDs found in the overflow area of the advertisement data (iOS only).
+  List<String> overflowServiceUUIDs;
+
+  /// A packet of data advertised by the device.
   AdvertisementData advertisementData;
 
+  /// Deserializes [ScanResult] from JSON.
   ScanResult.fromJson(Map<String, dynamic> json, ManagerForPeripheral manager)
```
mikolak: Private API scheduled for refactor

lib/scan_result.dart
```diff
@@ -32,14 +44,31 @@ class ScanResult {
         advertisementData = AdvertisementData._fromJson(json);
 }
 
+/// A container for BLE advertisement data with convenient accessors for
+/// common items.
```
mikolak: ?

lib/scan_result.dart
```diff
@@ -32,14 +44,31 @@ class ScanResult {
...
+
+  /// A dictionary that contains service-specific advertisement data.
   Map<String, Uint8List> serviceData;
+
+  /// A list of service UUIDs.
   List<String> serviceUUIDs;
+
+  /// The local name of the device.
+  ///
+  /// It is usually the same as [Peripheral.name] but sometimes one of them or
```
mikolak: Not so sure about that.

lib/scan_result.dart
```diff
@@ -32,14 +44,31 @@ class ScanResult {
...
+  /// both can be `null`.
   String localName;
+
+  /// The transmit power of the device.
   int txPowerLevel;
+
+  /// A list of solicited service UUIDs.
   List<String> solicitedServiceUUIDs;
 
+  /// Deserializes [AdvertisementData] from JSON.
```
mikolak: Private API scheduled for refactor

lib/characteristic.dart
```diff
@@ -11,16 +11,35 @@ abstract class _CharacteristicMetadata {
   static const String value = "value";
 }
 
+/// A characteristic of a local peripheral's [Service].
+///
+/// It contains a single value and any number of [Descriptor]s describing that
```
mikolak: Descriptors describe the Characteristic itself, not its value, I think. They're kind of metadata, I guess?

lib/characteristic.dart
```diff
@@ -56,16 +82,19 @@ class Characteristic extends InternalCharacteristic {
         transactionId ?? TransactionIdGenerator.getNextId(),
       );
 
+  /// Returns a [Stream] of values emitted by this characteristic.
```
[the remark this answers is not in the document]
lukerayman: According to [PREFER using “this”](https://dart.dev/guides/language/effective-dart/documentation#prefer-using-this-instead-of-the-to-refer-to-a-members-instance) rule I think we should keep `this characteristic` instead of `the characteristic`.

lib/scan_result.dart
lukerayman: I checked it on many devices. I think it's worth to mention that there are two places where you can find the name of the device. So if one of them is null you can always try to use the other one, so you have better chance to display some name instead of _some device_.

lib/scan_result.dart
lukerayman: Isn't it true? 😅

lib/characteristic.dart
lukerayman: To be honest I found it in CoreBluetooth documentation. I assumed that Apple is not wrong. But maybe... 🤔

lib/scan_result.dart
lukerayman: I added `.` to the end of this sentence. :)

lib/peripheral.dart
```diff
@@ -33,27 +49,49 @@ class Peripheral {
...
       _manager.characteristics(this, servicedUuid);
 
+  /// Reads RSSI for the device.
+  ///
+  /// Optional [transactionId] could be used to cancel operation.
   Future<int> rssi({String transactionId}) =>
       _manager.rssi(this, transactionId ?? TransactionIdGenerator.getNextId());
 
+  /// Request new MTU value for this device. This function currently is not
+  /// doing anything on iOS platform as MTU exchange is done automatically.
```
dariuszseweryn: Add info that if `requestMtu` was set during connection then it will also do nothing?

lib/peripheral.dart
```diff
@@ -33,27 +49,49 @@ class Peripheral {
...
+  /// Given [mtu] will be negotiated and the actual value is returned as future.
+  /// Optional [transactionId] could be used to cancel operation.
   Future<int> requestMtu(int mtu, {String transactionId}) =>
       _manager.requestMtu(
           this, mtu, transactionId ?? TransactionIdGenerator.getNextId());
 
+  /// Read characteristic value.
+  ///
+  /// Returns value of characteristic with [characteristicUUID] for service with
+  /// [serviceUUID]. Optional [transactionId] could be used to cancel operation.
```
dariuszseweryn: Should we inform the user that they always would need to discover services before? Same for write characteristic value?

lib/scan_result.dart
dariuszseweryn:
```suggestion
/// Type of objects returned by [BleManager] during scanning.
```

dariuszseweryn:
```suggestion
/// Type of objects returned by [BleManager] during scanning.
```

lib/scan_result.dart
```diff
@@ -15,14 +15,26 @@ abstract class _ScanResultMetadata {
...
 }
 
+/// Type of items returned by [BleManager] during scanning.
 class ScanResult {
   Peripheral peripheral;
+
+  /// Signal strength of the device in dBm.
   int rssi;
+
+  /// The size of the data unit within an ATT packet.
```
dariuszseweryn: It has no sense during a scan. It should be marked as `deprecated` on the spot

lib/scan_result.dart
dariuszseweryn: Also maybe add info that most of the properties here are optional?

lib/service.dart
```diff
@@ -20,9 +27,16 @@ class Service extends InternalService {
...
+  /// Returns a list of [Characteristic]s of this service.
   Future<List<Characteristic>> characteristics() =>
       _manager.characteristicsForService(this);
 
+  /// Writes the value in [bytes] of a [Characteristic] identified by
+  /// [characteristicUUID].
+  ///
+  /// It returns a [Future] that completes with the [Characteristic]. Only
+  /// [Characteristic] where [Characteristic.isWritableWithResponse] or
+  /// [Characteristic.isWritableWithoutResponse] is `true` can be written.
```
dariuszseweryn: appropriate value in `withResposne` parameter has to be set

lib/characteristic.dart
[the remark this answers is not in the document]
lukerayman: Also, in my opinion the word `characteristic` should be written in lower case if it's not linked to `Characteristic` class. I found that in Apple's docs for Core Bluetooth.

lib/peripheral.dart
```diff
@@ -4,7 +4,7 @@ abstract class _PeripheralMetadata {
   static const name = "name";
   static const identifier = "id";
 }
-
+/// Device instance represents BLE peripheral
```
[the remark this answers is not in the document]
lukerayman: Also `.` is missing at the and of this sentence.

lib/peripheral.dart
```diff
@@ -4,19 +4,38 @@ abstract class _PeripheralMetadata {
...
+  /// Creates peripheral from [json] and initializes it with [manager]
   Peripheral.fromJson(Map<String, dynamic> json, ManagerForPeripheral manager)
       : _manager = manager,
         name = json[_PeripheralMetadata.name],
         identifier = json[_PeripheralMetadata.identifier];
 
+  /// Connects to the Device.
+  ///
+  /// Optional [isAutoConnect] controls whether to directly connect to the
+  /// remote device (false) or to automatically connect as soon as the
```
lukerayman:
```suggestion
  /// remote device (`false`) or to automatically connect as soon as the
```

lib/peripheral.dart
```diff
@@ -4,19 +4,38 @@ abstract class _PeripheralMetadata {
...
+  /// remote device becomes available (true). [Android only]
```
lukerayman:
```suggestion
  /// remote device becomes available (Android only).
```
As `[]` formats text as code we should not use it for regular text.

lib/peripheral.dart
```diff
@@ -4,19 +4,38 @@ abstract class _PeripheralMetadata {
...
+  /// Optional [requestMtu] size will be negotiated to this value. It is not
+  /// guaranteed to get it after connection is successful. 0 passed as
+  /// Optional [requestMtu] means that the MTU will not be negotiated
```
lukerayman:
```suggestion
  /// optional [requestMtu] means that the MTU will not be negotiated.
```

lib/peripheral.dart
```diff
@@ -4,19 +4,38 @@ abstract class _PeripheralMetadata {
...
+  /// Passing true as [refreshGatt] leads reset services cache. This option may
```
lukerayman:
```suggestion
  /// Passing `true` as [refreshGatt] leads reset services cache. This option may
```

lib/peripheral.dart
```diff
@@ -4,19 +4,38 @@ abstract class _PeripheralMetadata {
...
+  /// be useful when a peripheral's firmware was updated and it's
```
lukerayman:
```suggestion
  /// be useful when a peripheral's firmware was updated and its
```

lib/peripheral.dart
```diff
@@ -4,19 +4,38 @@ abstract class _PeripheralMetadata {
...
+  /// services/characteristics were added/removed/altered. [Android only]
```
lukerayman:
```suggestion
  /// services/characteristics were added/removed/altered (Android only).
```
As `[...]` formats text as code we should not use it for regular text.

lib/peripheral.dart
```diff
@@ -4,19 +4,38 @@ abstract class _PeripheralMetadata {
...
+  /// Optional [timeout] is used to define time after connection is
+  /// automatically timed out. In case of race condition were connection
+  /// is established right after timeout event, device will be disconnected
+  /// immediately. Time out may happen earlier then specified due to OS
```
lukerayman:
```suggestion
  /// immediately. Timeout may happen earlier then specified due to OS
```

lib/peripheral.dart
```diff
@@ -28,32 +47,57 @@ class Peripheral {
           refreshGatt: refreshGatt,
           timeout: timeout);
 
+  /// Observe connection state
+  ///
+  /// Returns stream of [PeripheralConnectionState]
```
lukerayman: I think it's too short to break it into two paragraphs. I would keep it in one line. And dots at the end of both lines are missing.

lib/peripheral.dart
```diff
@@ -33,27 +49,49 @@ class Peripheral {
       _manager.observePeripheralConnectionState(
           identifier, emitCurrentValue, completeOnDisconnect);
 
+  /// Check connection state of the Device.
```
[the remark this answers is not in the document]
lukerayman: Missing `.` at the end of this sentence.

lib/peripheral.dart
```diff
@@ -33,27 +49,49 @@ class Peripheral {
...
   Future<bool> isConnected() => _manager.isPeripheralConnected(identifier);
 
+  /// Disconnects from Device if it's connected or cancels pending connection.
```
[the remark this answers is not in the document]
lukerayman: I would also add `this` before `peripheral` word.

lib/peripheral.dart
```diff
@@ -28,32 +47,57 @@ class Peripheral {
...
   Future<void> disconnectOrCancelConnection() =>
       _manager.disconnectOrCancelPeripheralConnection(identifier);
 
+  /// Discovers all Services, Characteristics and Descriptors for Device.
```
lukerayman:
```suggestion
  /// Discovers all services, characteristics and descriptors for this peripheral.
```
Is we don't link to classes I think it's better to have this in lowercase.

lib/peripheral.dart
```diff
@@ -28,32 +47,57 @@ class Peripheral {
...
+  ///
+  /// Optional [transactionId] could be used to cancel operation.
   Future<void> discoverAllServicesAndCharacteristics({String transactionId}) =>
       _manager.discoverAllServicesAndCharacteristics(
           this, transactionId ?? TransactionIdGenerator.getNextId());
 
+  /// List of discovered Services for the Device.
```
lukerayman:
```suggestion
  /// A list of discovered [Service]s for this peripheral.
```
I found in Flutter docs that it's ok to link to class and add `s` for plural.

lib/peripheral.dart
```diff
@@ -28,32 +47,57 @@ class Peripheral {
...
   Future<List<Service>> services() => _manager.services(this);
 
+  /// List of discovered Characteristics for given Service.
```
lukerayman:
```suggestion
  /// A list of discovered [Characteristic]s of a [Service] identified by [servicedUuid].
```

lib/peripheral.dart
```diff
@@ -28,32 +47,57 @@ class Peripheral {
...
   Future<List<Characteristic>> characteristics(String servicedUuid) =>
       _manager.characteristics(this, servicedUuid);
 
+  /// Reads RSSI for the device.
+  ///
+  /// Optional [transactionId] could be used to cancel operation.
   Future<int> rssi({String transactionId}) =>
       _manager.rssi(this, transactionId ?? TransactionIdGenerator.getNextId());
 
+  /// Request new MTU value for this device. This function currently is not
```
lukerayman:
```suggestion
  /// Request new MTU value for this peripheral. This function currently is not
```
Since we're in peripheral class I think it's more obvious when we use the same naming in doc.

lib/peripheral.dart
```diff
@@ -28,32 +47,57 @@ class Peripheral {
...
+  /// doing anything on iOS platform as MTU exchange is done automatically.
+  ///
+  /// Given [mtu] will be negotiated and the actual value is returned as future.
```
lukerayman:
```suggestion
  /// Given [mtu] will be negotiated and the actual value is returned as a [Future].
```

lib/peripheral.dart
```diff
@@ -28,32 +47,57 @@ class Peripheral {
...
+  /// Optional [transactionId] could be used to cancel operation.
   Future<int> requestMtu(int mtu, {String transactionId}) =>
       _manager.requestMtu(
           this, mtu, transactionId ?? TransactionIdGenerator.getNextId());
 
+  /// Read characteristic value.
```
lukerayman:
```suggestion
  /// Reads the value of a [Characteristic] identified by [characteristicUUID].
```

lib/peripheral.dart
```diff
@@ -66,6 +110,10 @@ class Peripheral {
         transactionId ?? TransactionIdGenerator.getNextId(),
       );
 
+  /// Write Characteristic value.
```
lukerayman:
```suggestion
  /// Writes the value of a [Characteristic] identified [characteristicUUID].
```

According to [third-person rule](https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs).

lib/peripheral.dart
```diff
@@ -82,12 +130,22 @@ class Peripheral {
         transactionId ?? TransactionIdGenerator.getNextId(),
       );
 
+  /// Fetch list of Descriptors
```
lukerayman: Should use third-person form.
```suggestion
  /// Fetches a list of [Descriptor]s.
```

lib/peripheral.dart
lukerayman: When you generate docs with `dartdocs` the you can see that only the first line is visible and to see the next paragraph, you have to click `[...]` which opens a new page. I think it's too much for such a short message. :D

lib/scan_result.dart
mikolak: It is, it just doesn't have much value. 
Would `A container for data with convenient accessors for common items.` be helpful to you if you saw it in a library's documentation?

lib/scan_result.dart
mikolak: Should be removed on the spot! I'll do it today!

lib/ble_manager.dart
```diff
@@ -1,9 +1,35 @@
...
+///
+/// Initialising/deinitialising native clients:
+/// ```dart
+/// BleManager bleManager = BleManager();
+/// await bleManager.createClient(); //ready to go!
+/// //your BLE logic
+/// bleManager.destroyClient(); //remember to release native resources when they're no longer needed
+/// ```
+///
+/// Obtaining [Peripheral]:
```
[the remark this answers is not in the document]
mikolak: We don't have it yet.

lib/peripheral.dart
```diff
@@ -12,6 +12,9 @@ class Peripheral {
   String name;
   String identifier;
 
+  /// Creates Peripheral from JSON
+  ///
+  /// Creates peripheral from [json] and initializes it with [manager]
```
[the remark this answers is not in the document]
mikolak: Yep.

lib/peripheral.dart
mikolak: [] is actually for linking, formatting as code is done as usual in MD using backticks

lib/peripheral.dart
mikolak: It will actually result in an error :D

PR #444, opened 2020-03-06
Characteristic unit tests

bartoszwilk:


test/characteristic_test.dart
```diff
@@ -0,0 +1,356 @@
...
+      descriptorGenerator.create(seed, characteristic);
+
+  tearDown(() {
+    [
+      peripheral,
+      managerForCharacteristic,
+    ].forEach(clearInteractions);
+  });
+
+  test("descriptors returns descriptors provided by manager", () async {
```
mikolak:
```suggestion
  test("descriptors returns a list of descriptors provided by manager", () async {
```

test/characteristic_test.dart
```diff
@@ -0,0 +1,356 @@
...
+
+    //then
+    expect(value, equals(Uint8List.fromList([1, 2, 3, 4])));
+  });
+
+  test(
+      "read invokes manager with expected params when transactionId is specified",
+      () {
+    //when
+    characteristic.read(transactionId: "123");
```
mikolak:
```suggestion
    characteristic.read(transactionId: "a123");
```

test/characteristic_test.dart
```diff
@@ -0,0 +1,356 @@
...
+
+    //then
+    verify(
+      managerForCharacteristic.readCharacteristicForIdentifier(
+          any, characteristic, "123"),
+    );
+  });
+
+  test(
+      "read invokes manager with expected params when transactionId is not specified",
```
mikolak:
```suggestion
      "read generates transactionId when it is not specified",
```

test/characteristic_test.dart
```diff
@@ -0,0 +1,356 @@
...
+
+    //then
+    verify(
+      managerForCharacteristic.readCharacteristicForIdentifier(
+          any, characteristic, argThat(isNotNull)),
+    );
+  });
+
+  test(
+      "read invokes manager with unique transactionId when transactionId is not specified",
```
mikolak:
```suggestion
      "read generates unique transactionId for each operation",
```

test/characteristic_test.dart
```diff
@@ -0,0 +1,356 @@
...
+                any,
+                characteristic,
+                Uint8List.fromList([1, 2, 3, 4]),
+                false,
+                captureThat(isNotNull)))
+        .captured;
+    expect(transactionIds[0], isNot(equals(transactionIds[1])));
+  });
+
+  test("monitor streams expected values", () {
```
mikolak:
```suggestion
  test("monitor emits expected values", () {
```

mikolak:
@bartoszwilk ?

test/characteristic_test.dart
bartoszwilk: Is it only about not using integers as transaction ids as they should be provided in a word-like form?

mikolak: Integers are reserved for the library and the user should not use them. Since tests might be used as a reference point for using the library, they should be written in a way that can be copy-pasted by new users.

PR #459, opened 2020-05-07
Release 2.2.4

mikolak:
* Fix issue where `withResponse` argument was always true when writing to a characteristic on iOS
* Remove unnecessary file that was interfering with analyzer


test/descriptor_test.dart
```diff
@@ -0,0 +1,118 @@
...
+
+import 'package:flutter_ble_lib/flutter_ble_lib.dart';
+import 'package:flutter_ble_lib/src/_managers_for_classes.dart';
+import 'package:mockito/mockito.dart';
+import 'package:test/test.dart';
+
+import 'mock/mocks.dart';
+import 'test_util/descriptor_generator.dart';
+
+void main() {
```
pawelByszewski, line 12: <3

PR #491, opened 2020-07-06
Release 2.2.6

dariuszseweryn:


.gitignore
```diff
@@ -13,4 +13,5 @@ res/values/strings_en.arb
 lib/generated/
 example/lib/generated/
 example/.flutter-plugins-dependencies
-.dart_tool/
\ No newline at end of file
+.dart_tool/
+example/ios/Podfile.lock
```
lukerayman, line 17: Why to ignore Podfile.lock? I think it's good to have it in the repo.

mikolak: No, since we're using pubspec and pubspec.lock for dependencies. Podfile is regenerated by Flutter.

PR #494, opened 2020-07-16
Added pedantic lib for better static analysis

dariuszseweryn:
Fixed (what I hope) are the biggest static analysis issues on pub.dev.

### DIFF
--- a/lib/ble_manager.dart
+++ b/lib/ble_manager.dart
@@ -34,9 +34,7 @@ abstract class BleManager {
   static BleManager _instance;
 
   factory BleManager() {
-    if (_instance == null) {
-      _instance = InternalBleManager();
-    }
+    _instance ??= InternalBleManager();
 
     return _instance;
   }
--- a/lib/scan_result.dart
+++ b/lib/scan_result.dart
@@ -1,8 +1,8 @@
 part of flutter_ble_lib;
 
 abstract class _ScanResultMetadata {
-  static const String id = "id";
-  static const String name = "name";
+//  static const String id = "id";
+//  static const String name = "name";
   static const String rssi = "rssi";
   static const String manufacturerData = "manufacturerData";
   static const String serviceData = "serviceData";
--- a/lib/scan_result.dart
+++ b/lib/scan_result.dart
@@ -1,8 +1,6 @@
 part of flutter_ble_lib;
 
 abstract class _ScanResultMetadata {
-//  static const String id = "id";
-//  static const String name = "name";
   static const String rssi = "rssi";
   static const String manufacturerData = "manufacturerData";
   static const String serviceData = "serviceData";
--- a/lib/src/_internal.dart
+++ b/lib/src/_internal.dart
@@ -12,7 +12,6 @@ import 'package:flutter_ble_lib/src/_constants.dart';
 import 'package:flutter_ble_lib/src/_containers.dart';
 import 'package:flutter_ble_lib/src/util/_transaction_id_generator.dart';
 import 'package:flutter_ble_lib/src/util/_transformers.dart';
-import 'package:pedantic/pedantic.dart';
 
 import '_managers_for_classes.dart';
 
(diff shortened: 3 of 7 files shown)


lib/ble_manager.dart
```diff
@@ -34,9 +34,7 @@ abstract class BleManager {
   static BleManager _instance;
 
   factory BleManager() {
-    if (_instance == null) {
-      _instance = InternalBleManager();
-    }
+    _instance ??= InternalBleManager();
```
pawelByszewski, line 37: wow!!

lib/scan_result.dart
```diff
@@ -1,8 +1,8 @@
 part of flutter_ble_lib;
 
 abstract class _ScanResultMetadata {
-  static const String id = "id";
-  static const String name = "name";
+//  static const String id = "id";
```
pawelByszewski, line 4: shouldn't the fields be removed?

dariuszseweryn: I dunno. I just wanted to get rid of [these static analysis](https://pub.dev/packages/flutter_ble_lib/score) issues. I lack context.

lib/ble_manager.dart
dariuszseweryn: Use `pedantic` from the beginning and it can show you quite some tricks 👍

lib/src/bridge/device_connection_mixin.dart
```diff
@@ -39,16 +39,16 @@ mixin DeviceConnectionMixin on FlutterBLE {
           return PeripheralConnectionState.disconnecting;
         default:
           throw FormatException(
-              "Unrecognized value of device connection state. Value: $connectionStateString");
+              'Unrecognized value of device connection state. Value: $connectionStateString');
       }
     });
 
-    _methodChannel.invokeMethod(
+    unawaited(_methodChannel.invokeMethod(
```
dariuszseweryn, line 46: @mikolak Especially your attention should be here. I am not sure if it was "unawaited" intentionally and I lack a bit to be sure now.

lib/scan_result.dart
mikolak: They should be removed, they're used for passing peripheral information, but it's deserialized using `Peripheral.fromJson` and its own metadata.

lib/src/bridge/device_connection_mixin.dart
mikolak: IIRC the values might start coming before the future returns. We were trying to make the call non-blocking, so that first values could come already. This was just as we started working on the library and we wanted to avoid stream controllers by using yield*. 

This can be improved by declaring a `StreamController` and passing the invoking of the method to its `onListen` instead, same construction as in monitoring characteristics.

lib/src/_internal.dart
```diff
@@ -12,6 +12,7 @@ import 'package:flutter_ble_lib/src/_constants.dart';
 import 'package:flutter_ble_lib/src/_containers.dart';
 import 'package:flutter_ble_lib/src/util/_transaction_id_generator.dart';
 import 'package:flutter_ble_lib/src/util/_transformers.dart';
+import 'package:pedantic/pedantic.dart';
```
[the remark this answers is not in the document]
dariuszseweryn: It contains "unawaited" function.

lib/src/bridge/device_connection_mixin.dart
```diff
@@ -14,13 +14,30 @@ mixin DeviceConnectionMixin on FlutterBLE {
...
   Stream<PeripheralConnectionState> observePeripheralConnectionState(
-      String identifier, bool emitCurrentValue) async* {
-    yield* _peripheralConnectionStateChanges
+      String identifier, bool emitCurrentValue) {
+    var controller = StreamController<String>(
+      onListen: () => _methodChannel
+          .invokeMethod(MethodName.observeConnectionState, <String, dynamic>{
+        ArgumentName.deviceIdentifier: identifier,
+        ArgumentName.emitCurrentValue: emitCurrentValue,
+      }).catchError(
```
mikolak, line 31:
```suggestion
      },).catchError(
```
and then formatter :)

PR #502, opened 2020-07-28
Add unsafe way of creating peripherals

### LINKED ISSUE #295: Connect to a known peripheral without scanning
`bleManager.connectToPeripheral(String peripheralId)`

mikolak:
Resolves #295

### DIFF
--- a/lib/ble_manager.dart
+++ b/lib/ble_manager.dart
@@ -137,6 +137,11 @@ abstract class BleManager {
   ///
   /// If [serviceUUIDs] is empty, this will return an empty list.
   Future<List<Peripheral>> connectedPeripherals(List<String> serviceUUIDs);
+
+  /// Creates a peripheral which may not exist. Since the [peripheralId] might
+  /// be a UUID or a MAC address, depending on the platform, its format is
+  /// not validated.
+  Peripheral createUnsafePeripheral(String peripheralId, {String name});
 }
 
 /// State of the Bluetooth Adapter.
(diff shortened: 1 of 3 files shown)


lib/src/internal_ble_manager.dart
```diff
@@ -62,6 +62,16 @@ class InternalBleManager
   @override
   Future<void> stopPeripheralScan() => _bleLib.stopDeviceScan();
 
+  @override
+  Peripheral createUnsafePeripheral(String peripheralId, {String name}) {
+    const nameField = 'name';
```
dariuszseweryn, line 67: I wish we had those constants described in a single place. Maybe protobuf?

lib/ble_manager.dart
```diff
@@ -137,6 +137,11 @@ abstract class BleManager {
   ///
   /// If [serviceUUIDs] is empty, this will return an empty list.
   Future<List<Peripheral>> connectedPeripherals(List<String> serviceUUIDs);
+
+  /// Creates a peripheral which may not exist. Since the [peripheralId] might
+  /// be a UUID or a MAC address, depending on the platform, its format is
+  /// not validated.
+  Peripheral createUnsafePeripheral(String peripheralId, {String name});
```
dariuszseweryn, line 144: Maybe add an information that 'peripheralId' passed here is meant to be only the `peripheralId` of a previously scanned peripheral on this particular device — as a general rule 'peripheralId' is valid only on the same device it was scanned.

VinciShark:
I test this function with:
```
var peripheral = bleManager.createUnsafePeripheral({MY_ID_HERE});
await peripheral.connect();
bool connected = await peripheral.isConnected();
print("!!!!" + connected.toString());
```
Works fine! LGTM

PR #547, opened 2020-11-16
Fix emitting current connection state on iOS

### LINKED ISSUE #540: observeConnectionState(emitCurrentValue: true) emits connected state for disconnected device
I found a bug in 2.3.0 version of the library. It can be reproduces with the sample app using iOS device

Reproduction steps:
1. Connect to any device
2. Turn off the device
3. The app should report the device as disconnected
4. Try to connect to the device
5. The app reports the device as connected

Also can be reproduced like this:
1. Scan for devices
2. Open device details
3. Turn off the device
4. Try to connect
5. The app reports the device as connected

mikolak:
For reasons unknown casting `BOOL: NO` to `BOOL` makes it a `YES`.

Fixes #540

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 2.3.1
 
 * Fix connection timeout on iOS
+* Fix emitting current connection state on iOS
 
 ## 2.3.0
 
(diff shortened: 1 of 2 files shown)


dariuszseweryn:
I guess it may be that `(id)` is a pointer to a `BOOL` but when you cast that pointer to `BOOL` it becomes `YES` as the pointer is not a `nil` reference. 🤷‍♂️

dariuszseweryn (approved):

PR #548, opened 2020-11-19
Release v2.3.1

mikolak:
## 2.3.1

* Fix connection timeout on iOS
* Fix emitting current connection state on iOS
* Update MBA to 0.1.7:
    * Fix lack of disconnection event on iOS when connection fails to be established
    * Fix not all errors being passed through onError callbacks on Android (thanks, @eliaslecomte)


mikolak (commented):
✅

PR #557, opened 2020-12-11
Prepare release of 3.0.0-beta

mikolak:
## 3.0.0-beta

* **BREAKING** Migrate Android implementation to RxAndroidBle2
    This might require the user to implement a global error handler https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling
    See more: https://github.com/Polidea/RxAndroidBle/wiki/FAQ:-UndeliverableException


dariuszseweryn (approved):
I actually think whether it is a breaking change or would it be more suited to be 2.4.0-beta. Technically it does not need any changes to work.

mikolak:
Yeah, but on the other hand there's a minimal chance it will cause a fatal crash if the error handler is not added.

PR #584, opened 2021-03-24
Update json scan_result parsing

sjoerd0301:
iOS has overflow uuids, the list containing these items is a `List<dynamic>` which couldn't be cast to a `List<String>`


lib/scan_result.dart
```diff
@@ -32,7 +32,7 @@ class ScanResult {
       : peripheral = Peripheral.fromJson(json, manager),
         rssi = json[_ScanResultMetadata.rssi],
         isConnectable = json[_ScanResultMetadata.isConnectable],
-        overflowServiceUuids = json[_ScanResultMetadata.overflowServiceUuids],
+        overflowServiceUuids = _mapToListOfStringsOrNull(json[_ScanResultMetadata.overflowServiceUuids]),
```
mikolak: 👍

lib/scan_result.dart
```diff
@@ -78,12 +78,12 @@ class AdvertisementData {
...
+    } else {
       return null;
+    }
   }
-
-  static List<String> _mapToListOfStringsOrNull(List<dynamic> values) =>
-      values?.cast();
 }
+
+List<String> _mapToListOfStringsOrNull(List<dynamic> values) => (values ?? []).cast();
```
mikolak: Why the change of behaviour? Most of those properties might be an array or null on native and that's what this function did - either map dynamics from the list to strings or preserve the null.

sjoerd0301: Wouldn't it be a better idea to, with recent nullsafety changes in mind, return an empty list instead of null?
Otherwise it would be no problem to just continue with the `values?.cast()`

mikolak: Null is a more definitive indication that there are no such values retrieved. I like the idea of eliminating null, but I'd have to examine the rest of this file and analyze if this could be breaking or not. I'll try to do that tomorrow.

PR #586, opened 2021-04-01
Nullsafety

### LINKED ISSUE #556: Null-Safety migration
Do you plan migrate to null-safety? Next Dart(2.12) will be with stable and by default null-safety.

okocsis:
Hey, guys! I've made a lot of effort to port the dart codebase to a nullsafe environment. I would appreciate someone taking a look at it.

known issues:
-not all of the test are compiling, although I've tried to make it work I've even ported some of the mocks to the new generated mocking scheme. FIXME: I think many of the test should be retired, in favour of general nullsafety. Please help me on that.
-there was a bleadapter lifecycle fiasco, because the DevicesBloc destroyed the client on dispose(), making the internal adapter unreachable, after going to detail view. Although I have attempted to fix this by just removing the destroy client call from DevicesBloc.dispose() but I'm just not certain this was the right solution in terms of good design.

Thanks for taking the time!

Fixes #556


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "editor.rulers": [
```
mikolak: Add the file to .gitignore and remove it from PR, please :)

.travis.yml
```diff
@@ -55,6 +55,11 @@ _ios_job_template: &ios_job_template
     - export PATH=`pwd`/flutter/bin:`pwd`/flutter/bin/cache/dart-sdk/bin:$PATH
   script:
     - cd example
+    - flutter pub upgrade
```
mikolak: Example shouldn't upgrade it's dependencies on CI only, this will quickly make it even more stale than it is now. ;)

lib/ble_manager.dart
```diff
@@ -147,7 +151,7 @@ abstract class BleManager {
   /// On Android [peripheralId] scanned on one  device may or may not be
   /// recognized on a different Android device depending on peripheral’s
   /// implementation and changes in future OS releases.
-  Peripheral createUnsafePeripheral(String peripheralId, {String name});
+  Peripheral createUnsafePeripheral(String peripheralId, {String? name});
```
mikolak, line 157: Does an optional argument has to be explicitly typed as nullable? There's no point in explicitly passing a null to an optional argument if that's not the case.

lib/characteristic.dart
```diff
@@ -20,7 +20,7 @@ class Characteristic extends InternalCharacteristic {
   /// The [Service] containing this characteristic.
   Service service;
```
mikolak: This should also be final, like the manager

lib/characteristic.dart
```diff
@@ -20,7 +20,7 @@ class Characteristic extends InternalCharacteristic {
...
 
-  ManagerForCharacteristic _manager;
+  final ManagerForCharacteristic _manager;
 
   /// The UUID of this characteristic.
   String uuid;
```
mikolak, line 26: Like above, also final 🙂

mikolak (commented):
Just some early feedback as I'm on 11/51 files - I'll finish it tomorrow. Looks nice so far, thanks for the PR!

lib/ble_manager.dart
okocsis: yes, that's what the new dart standard says, doesn't even let you "compile" it without it. 
So to sum up name needs to eighter:
- have a default value {String name = ""} (I'm down with that too)
- or be explicitly nullable, so if the caller omits it. it will be null by default implicitly, that needs to be nullchecked thus the type must be marked nullable in that case

lib/characteristic.dart
okocsis: yes, indeed, in fact, a lot of vars and fields could be final, I only changed the ones I touched, so the PR remains relatively clean. But I generally agree with the "prefer final" idiom everywhere. I think it will be cleaner to do that refactor in another PR imho. :)

.travis.yml
okocsis: gotcha, I was just playing around to make it pass somehow anyways ;)

.vscode/settings.json
okocsis: sure thing!

.travis.yml
okocsis: By the way, do you happen to have any better idea why it fails on CI environment, 
I don't see a reason for it not to work, but it doesn't seem to use the correct updated version of permission_handler package, (upgraded to have more sound nullsafety, also bugfixes were taken place since the previous version)
I'm on Xcode 12.4, flutter 2.0.4.

lib/characteristic.dart
okocsis: Do we want any of the fields of this object (POD) to be mutable? in other words, does the user ever want to alter any of the fields of a retrieved Characteristic object, if anything it only leads to inconsistency innit?

mikolak: I wrote about it since you've finaled things on descriptor, but not on Characteristic. What needs to be done is actually move all the implementation away from characteristic to internal characteristic and make Characteristic abstract with everything final, so that it's only an interface, but it's somewhere on my to-do. Either way no worries, these are just nitpicks and I'll go over the all the classes before release, you can safely ignore it in this PR.

.travis.yml
mikolak: Maybe CI is running outdated XCode?

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -228,13 +228,11 @@
 			);
 			inputPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-Runner/Pods-Runner-frameworks.sh",
-				"${PODS_ROOT}/../Flutter/Flutter.framework",
```
mikolak: Are changes here intended?

.travis.yml
okocsis: can easily be that, shall i ammend it or will address it?

lib/characteristic.dart
```diff
@@ -170,21 +168,22 @@ class Characteristic extends InternalCharacteristic {
 ///
 /// This type is created to support chaining of operations on the characteristic
 /// when it was first read from [Peripheral] or [Service].
-class CharacteristicWithValue extends Characteristic with WithValue {
+class CharacteristicWithValue extends Characteristic {
+  Uint8List value;
```
mikolak, line 172: Why move value from the mixin?

lib/error/ble_error.dart
```diff
@@ -13,23 +14,24 @@ abstract class _BleErrorMetadata {
...
+  final String? internalMessage;
 
   BleError.fromJson(Map<String, dynamic> json)
-      : errorCode = BleErrorCode(json[_BleErrorMetadata.errorCode]),
+      : errorCode = BleErrorCode(json[_BleErrorMetadata.errorCode] ?? 0),
         attErrorCode = json[_BleErrorMetadata.attErrorCode],
+        iosErrorCode = json[_BleErrorMetadata.iosErrorCode],
         androidErrorCode = json[_BleErrorMetadata.androidErrorCode],
-        reason = json[_BleErrorMetadata.reason],
+        reason = json[_BleErrorMetadata.reason] ?? "",
```
mikolak:
```suggestion
        reason = json[_BleErrorMetadata.reason] ?? "Reason not provided",
```

lib/scan_result.dart
```diff
@@ -14,49 +14,66 @@ abstract class _ScanResultMetadata {
...
-  Peripheral peripheral;
+  final Peripheral peripheral;
 
   /// Signal strength of the peripheral in dBm.
-  int rssi;
+  final int rssi;
 
   /// An indicator whether the peripheral is connectable (iOS only).
-  bool isConnectable;
+  late final bool isConnectable;
```
mikolak: Why `late`, as it's initialised in the initialiser list? As it's iOS-only it should be nullable I think.

lib/scan_result.dart
```diff
@@ -14,49 +14,66 @@ abstract class _ScanResultMetadata {
...
 
   /// A list of UUIDs found in the overflow area of the advertisement data (iOS only).
-  List<String> overflowServiceUuids;
+  late final List<String> overflowServiceUuids;
```
mikolak: Why `late`, as it's initialised in the initialiser list? As it's iOS-only it should be nullable I think.

lib/scan_result.dart
```diff
@@ -14,49 +14,66 @@ abstract class _ScanResultMetadata {
...
+    List<String>? overflowServiceUuids, 
+  }) : isConnectable = isConnectable ?? false,
+       overflowServiceUuids = overflowServiceUuids ?? <String>[];
+
+
+  factory ScanResult.fromJson(
+    Map<String, dynamic?> json, 
+    ManagerForPeripheral manager
+  ) {
+    assert(json[_ScanResultMetadata.rssi] is int);
```
mikolak, line 44: Assert only works in debug. 
I don't see the benefit of making it a factory constructor and adding an additional private construct. What's the motivation here?

lib/scan_result.dart
```diff
@@ -14,49 +14,66 @@ abstract class _ScanResultMetadata {
...
+    );
+  }
 }
 
 /// Data advertised by the [Peripheral]: power level, local name,
 /// manufacturer's data, advertised [Service]s
 class AdvertisementData {
   /// The manufacturer data of the peripheral.
-  Uint8List manufacturerData;
+  Uint8List? manufacturerData;
```
mikolak: Properties of AdvertisementData should be final as well, I guess.

lib/src/bridge/characteristics_mixin.dart
```diff
@@ -169,7 +175,7 @@ mixin CharacteristicsMixin on FlutterBLE {
     ).map((characteristicWithValue) => characteristicWithValue.value);
   }
 
-  Stream<CharacteristicWithValue> monitorCharacteristicForDevice(
+  Stream<CharacteristicWithValueAndTransactionId> monitorCharacteristicForDevice(
```
mikolak: The motivation here was that the transactionId on the object should be obscured from the consumers, as they expect either Characteristic or CharacteristicWithValue and the transaction id is only used in this file. (this also applies to the rest of the definitions in this file)

lib/src/bridge/characteristics_mixin.dart
```diff
@@ -245,25 +252,31 @@ mixin CharacteristicsMixin on FlutterBLE {
...
+        .then((_) => streamController.close());
 
     return streamController.stream;
   }
 
   CharacteristicWithValueAndTransactionId
       _parseCharacteristicWithValueWithTransactionIdResponse(
-          Peripheral peripheral, rawJsonValue) {
-    Map<String, dynamic> rootObject = jsonDecode(rawJsonValue);
+          Peripheral peripheral, dynamic? rawJsonValue) {
```
mikolak: rawJsonValue can be just typed as String, I think? We must have missed type annotation and for reasons unknown Dart didn't complain

lib/src/bridge/scanning_mixin.dart
```diff
@@ -22,11 +22,13 @@ mixin ScanningMixin on FlutterBLE {
     List<String> uuids,
     bool allowDuplicates,
   ) {
-    if (_scanEvents == null) {
-      _prepareScanEventsStream();
+    var scanEvents = _scanEvents;
+    if (scanEvents == null) {
+      scanEvents = _prepareScanEventsStream();
+      _scanEvents = scanEvents;
     }
```
mikolak:
```suggestion
    var scanEvents = _scanEvents ?? _prepareScanEventsStream();
    _scanEvents = scanEvents;
```

mikolak: Or even better: move the second assignment back to `_prepare...` and instead make it return either existing _scanEvents or the newly created, singleton style

test/src/bridge/lib_core_test.dart
```diff
@@ -9,201 +9,338 @@ import 'package:flutter_ble_lib/src/_internal.dart';
...
+      bleLib.monitorCharacteristicForIdentifier(peripheral, 123, mockTransId),
+      emitsInOrder([
+        emitsError(isInstanceOf<BleError>()),
+        emitsDone,
+      ])
+    );
+    await _emitPlatformError(cancellationErrorJson(mockTransId));
+
+    // await fut;
+    print("anyad");
```
mikolak: Remove, please

mikolak (changes requested):
Got a few questions and a change request, but other than that looks OK!

.travis.yml
mikolak: If you could, please. 🙂  This line: https://github.com/Polidea/FlutterBleLib/blob/develop/.travis.yml#L50

test/src/bridge/lib_core_test.dart
okocsis: :D ups... it was hard fixing tests, so... :D

example/ios/Runner.xcodeproj/project.pbxproj
okocsis: I'm not even sure that this file is meant to be tracked on purpose or not.. maybe we should untrack this too?

lib/characteristic.dart
okocsis: I didn't get the point of the mixin here since it's not used anywhere else. 
Do you plan to use polymorphism over the type WithValue ever? I couldn't imagine any use case like for example if there was a function with the signature: `void foo(WithValue value)` I'd understand, one may pass in anything that conforms to WithValue thus has a `value` filed.

Otherwise, to me, it was cleaner and more concise to just have the field there, as jumping across files all the time takes a lot of time, for imho no benefit.
But again I might be missing some clue here.

lib/error/ble_error.dart
okocsis: hm, good suggestion! much better with actual text! thx
fixing it

lib/scan_result.dart
okocsis: Late is there by mistake, i needed it at first but after refactoring the constructor into an init list it's not needed anymore, thanks to good catch! I'm going to look for all of these unnecessary late qualifiers.
iOS only:
Or it could be understood as default as an empty list on android, therefore not needing to branch the code for iOS/android anymore for the user. 
In other words, does being null carry any useful information for the user here? 
What difference does it make if it's null or it's an empty list of overflowIds semantically? Let's make the caller's life as easy as possible.
On the other hand, if there is a single use-case that the caller would branch on a null differently to an empty list it absolutely must be nullable!

lib/scan_result.dart
okocsis: right.. and a lots of helper var's on local scope too, I know this `var` thing is just a "bad habit" originated from JS and stuff, the default thing one should type in is `final foo = Bar()` and not `var foo = Bar()` or `Bar foo = Bar()` -- which is also a var -- one should only use var when they mean to reassign it later to some other object.

lib/scan_result.dart
okocsis: now, that's why it's great that you review this, ofc you are much more aware of the edge cases and behaviours of you own library.
What do you think we move it to an enum instead since it's 3.0 milestone anyways. 
Like `final Connectivity connectivity;` with 3 states:
```Connectivity.noInfo, Connectivity.notConnectable, Connectivity.Connectable```
I'm actually okay with the nullable version as well, but I'm also anxious, people would merge the cases of false and null creating bad logic accidently. It's a common mistake. Thus I generally wouldn't prefer a nullable `bool?` field anywhere.

As the semantic difference of null and false is often unclear and some libs also use them interchangeably, meaning they sloppily emit null instead of false again JS's fault.. the truthy falsy fiasco...

File: lib/scan_result.dart (hunk 11 above)
Comment (okocsis) [reply lib/scan_result.dart:44]: motivations:
-easier to mock a dummy ScanResult object, you just need a @visibleForTesting on a static method that calls this private constructor.
-more importantly, type checker catches _TyepeError-s at the right place
-if something is null out of the dict object it can be concisely handled there in the constructor
-if you want to avoid late qualifiers, the fields must be initialised in the initialiser list, the constructor body -- here factory body -- is too late, so I wanted to avoid to having to us `late` as semantically it's not really late is it?

File: lib/src/bridge/characteristics_mixin.dart (hunk 13 above)
Comment (okocsis) [reply lib/src/bridge/characteristics_mixin.dart]: that's fine, but the test case for some reason expected a CharacteristicWithValueAndTransactionId, and as this is only a static type here and the runtime type was correct ofc, the test still failed, so I played around trying to fix it, and it turned out it was something else, a mocking behaviour changed and that broke the test, so I guess I now can safely change it back and "it SHALL PASS"! ;)

File: lib/src/bridge/characteristics_mixin.dart (hunk 14 above)
Comment (okocsis) [reply lib/src/bridge/characteristics_mixin.dart]: for rust to complain you need to set static type analysis to strict or something, (I just read this 3day ago).
Anyways, I thought it was dynamic for a reason so I went on and made it dynamic explicitly.
dynamic is really JS mode... :D, drilling a huge hole into the typesystem..

File: lib/src/bridge/scanning_mixin.dart (hunk 15 above)
Comment (okocsis) [reply lib/src/bridge/scanning_mixin.dart]: Okay, so this is annoying me too, it looks bad but that's a trick to make dart actually leverage its nullcheck static capabilities..
Here is the stupid thing in Dart:
if you just check on `if (_scanEvent == null)` then it wouldn't autounwrap _scanEvent after the line it actually gets a statically non-nullable value on line 27 as you recommended, so even still you'd need to unwrap _scanEvent in the local scope like `foo(_scanEvent!)`which is ugly since it's guaranteed to be nonnull.
The trick is if you do the same but with a `local ` and not with an `object field`. 
like so: `var scanEvents = _scanEvents;`
then no manual unwrap is needed the compiler will know that the local var `scanEvents` has a value for sure while it's unsure about its ivar alias `_scanEvents`.

So to sum up you'll see a lot of boilerplate helper code to mitigate this behaviour of Dart, although it's only 2 extra lines each time, and eliminates the need for `force unwrapping` operators, which is semantically noisy and should be avoided!

File: .travis.yml (hunk 2 above)
Comment (okocsis) [reply .travis.yml]: will do

File: lib/src/bridge/characteristics_mixin.dart (hunk 14 above)
Comment (okocsis) [reply lib/src/bridge/characteristics_mixin.dart]: I moved the rawJsonValue's String type check to the caller side and made the signature type-safe `String rawJsonValue`.

**Comment by okocsis** [comment]
@mikolak 
Okay so, it seems the new xcode image did the trick for travis, now phase two passes as well, but still fails on phase three `package-analysis` as follows:
"package-analysis
 It seems that this action doesn't have the required permissions to call the GitHub API with the token you gave. This can occur if this repository is a fork, as in that case GitHub reduces the GITHUB_TOKEN's permissions for security reasons. Consequently, no report will be made on GitHub. Check this issue for more information: https://github.com/axel-op/dart-package-analyzer/issues/2"
 
 I don't know how to fix this one mate.

**Comment by mikolak** [comment]
Yeah, it's because it's a PR from a fork instead of a MR from this repo - don't mind.

File: lib/src/bridge/scanning_mixin.dart (hunk 15 above)
Comment (mikolak) [reply lib/src/bridge/scanning_mixin.dart]: Yeah, I get it, but if _prepareScanEventsStream() always returns non-null and it does the 
```dart
if (_scanEvents == null) { 
   // create a new stream and assign it to _scanEvents
 }
 
 return _scanEvents
 ```
 
you still get that behaviour, but you get rid of the assignment in a function that does not care about it.

File: example/ios/Runner.xcodeproj/project.pbxproj (hunk 6 above)
Comment (mikolak) [reply example/ios/Runner.xcodeproj/project.pbxproj]: I'm fairly certain it should be tracked, as it's the definition of project contents. I saw on my branch that integration must have changed and also got those changes automatically. Just wanted to make sure it wasn't by mistake. 🙂

File: lib/scan_result.dart (hunk 11 above)
Comment (mikolak) [reply lib/scan_result.dart:44]: But the original constructor didn't have a body and was public? The only thing changed is (apart from the assert) `overflowServiceUuids ?? <String>[]`.

Doesn't really matter, as this should be changed to abstract class in the future with the implementation moved to src/. Marking as resolved.

File: lib/scan_result.dart (hunk 9 above)
Comment (mikolak) [reply lib/scan_result.dart]: Enum sounds OK. I've slapped the milestone 3.0 on this, but I'm actually inclined to release this as 2.4.0, as I don't see any breaking changes, so I'll create a separate issue for the enum.

Comment (mikolak) [reply lib/scan_result.dart]: https://github.com/Polidea/FlutterBleLib/issues/588

File: lib/characteristic.dart (hunk 7 above)
Comment (mikolak) [reply lib/characteristic.dart:172]: Only benefit was that it was used for Characteristic and Descriptor. Not much of a value, I agree.

File: lib/scan_result.dart (hunk 10 above)
Comment (mikolak) [reply lib/scan_result.dart]: Null avoids the issues like `it's always empty` lacking the crucial detail - tested on Android. ;) Though it makes sense to put ease of use over potential unnecessary questions, let's keep it this way.

File: lib/src/bridge/characteristics_mixin.dart (hunk 13 above)
Comment (mikolak) [reply lib/src/bridge/characteristics_mixin.dart]: I think this is the last thing?

File: lib/src/bridge/scanning_mixin.dart [hunk 17]
```diff
@@ -40,8 +42,8 @@ mixin ScanningMixin on FlutterBLE {
     );
 
     streamController
-        .addStream(_scanEvents, cancelOnError: true)
```
okocsis, line 43: @mikolak the problem lies here. If I leave it _scanEvents then the compiler is just not clever enough to know that _scanEvents is actually autoUnwrappable, and I'll have to write: `.addStream(_scanEvents!, cancelOnError: true)`
It's not that force unwrap operator is ugly, it's discouraged to use for a reason:
- Imagine, if the code above changes any time in the future, changing the behaviour of the code in way that `_scanEvents` is no longer guaranteed to be `nonnull`.
Nothing is gonna warn you about the `!` force unwrap now being semantically incorrect, and crashing the software on edge cases.
- Whereas if you maintain a helper local var instead /more verbosely I know/, then this behaviour change on 43 will break the compilation of the code since the compiler will see that `scanEvents` local var can no longer be proven `nonnull` and thus it will be `nullable` and `addStream()` don't accept no nullable value by its signature, so it will not compile, which is a good thing. It can safeguard you from future mistakes even, warning you to add an extra null check or something.

So yeah, it may seem to be an unnecessary complication with the extra var and all, but it really for semanticly correct nullsafe coding. 
Ofc, I'm also happy with anything that achieves the same, correct me if I'm wrong but I don't see you suggested change achieving the same safety concern.
Or b.) no fucks given and I should just make it less verbose sacrificing nullsafety.
Up to you ;)

lib/src/bridge/scanning_mixin.dart
okocsis: it will have to be `return _scanEvents!` actually, see below, I'll explain more thoroughly at the point of usage of scanEvenets.

mikolak: I don't mind a helper var in `_prepareScanEventsStream()`, I just don't see the point of having a function return non-nullable if we still need to check for null and reassign in the consumer of said function.

okocsis: Okay gotcha, now _scanEvents is a lazy getter, and _prepareScanEventsStream is gone, this way caller don't need to check anything. I actually would have written it like this myself, just again didn't want to introduce too much change out of the scope of this PR ;)

lib/src/bridge/scanning_mixin.dart
```diff
@@ -1,19 +1,29 @@
...
 
 mixin ScanningMixin on FlutterBLE {
-  Stream<ScanResult>? _scanEvents;
-
-  Stream<ScanResult> _prepareScanEventsStream() {
-    return const EventChannel(ChannelName.scanningEvents)
-        .receiveBroadcastStream()
-        .handleError(
-          (errorJson) => throw BleError.fromJson(jsonDecode(errorJson.details)),
+  Stream<ScanResult>? __scanEvents;
```
mikolak: How about `_activeScanEvents` or `_currentScanStream`? I don't like having a variable discerned only by having one more dash in front - too easy to fool the eye, especially with IDE suggestion.

lib/src/bridge/scanning_mixin.dart
```diff
@@ -1,19 +1,29 @@
...
+  Stream<ScanResult> get _scanEvents {
+    var scanEvents = __scanEvents;
+    if (scanEvents == null) {
+      scanEvents = 
+        const EventChannel(
+          ChannelName.scanningEvents
+        ).receiveBroadcastStream(
+        ).handleError(
```
mikolak: Has the formatter been run here? It looks weird.

mikolak (changes requested):

lib/src/bridge/scanning_mixin.dart
okocsis: it's a style, it's different to your style, to me it's easier to overlook function chaining like this because `).foo()`easier to catch for the eye then simply `.foo()` wich as you start to read it can also be an enum case with a leading dot /not in dart ofc/.. I don't mind changing it to just leading dot style, but it already wasn't consistent anyways.

I do not like autoformatters (or i haven't found one that actually works well)

lib/src/bridge/scanning_mixin.dart
okocsis: the goal was to make it super-duper private, and to denote that it wasn't meant to be used from the class scope only from the scope of the actual interface being:
`get _scanEvents` getter and `_resetScanEvents()` to indirectly interface that state variable in the class.
`_activeScanEvents` works too ofc. ;)

lib/src/bridge/characteristics_mixin.dart
```diff
@@ -66,11 +77,16 @@ mixin CharacteristicsMixin on FlutterBLE {
...
+            if (rawValue is String) {
+              rawJsonValue = rawValue;
+            }
+            return _parseCharacteristicWithValueWithTransactionIdResponse(
+              peripheral, 
+              rawJsonValue
+            );
+          });
 
   Future<void> writeCharacteristicForIdentifier(
```
okocsis, line 91: @mikolak 
OFF-TOPIC:
it looks like the response is swallowed here and it returns void instead of `Characteristic? response`
is this intentional?

Same in other write methods where the response is optionally available, all those should return `Future<CharacteristicsWithValue?>` as opposed to `Future<Characteristics>` right?

Or am I overlooking something? 🤔

okocsis: the last commit is to mitigate this. 
Pls recommend a better exception to throw on null value erroneously coming from native.
Thanks!

lib/src/bridge/scanning_mixin.dart
mikolak: pub.dev cuts points for code not formatted using the dartfmt. Like it or not, it's mandatory. 🙂

lib/src/bridge/bluetooth_state_mixin.dart
```diff
@@ -37,7 +39,7 @@ mixin BluetoothStateMixin on FlutterBLE {
     yield* _adapterStateChanges.map(_mapToBluetoothState);
   }
 
-  BluetoothState _mapToBluetoothState(dynamic rawValue) {
+  BluetoothState _mapToBluetoothState(String? rawValue) {
```
mikolak: Unless Dart complains, this should be typed as non-nullable. If for any reason a null came from the native side, then it's an error and should not be swept under the rug.

lib/service.dart
```diff
@@ -37,7 +37,7 @@ class Service extends InternalService {
   /// [Characteristic.isWritableWithResponse] or
   /// [Characteristic.isWritableWithoutResponse] is `true` and
   /// [withResponse] is specified accordingly can be written to.
-  Future<Characteristic> writeCharacteristic(
+  Future<Characteristic?> writeCharacteristic(
```
mikolak: This is not nullable, it's either characteristic or error

lib/src/_managers_for_classes.dart
```diff
@@ -52,7 +52,7 @@ abstract class ManagerForPeripheral {
     String transactionId,
   );
 
-  Future<Characteristic> writeCharacteristicForDevice(
+  Future<Characteristic?> writeCharacteristicForDevice(
```
mikolak: This is not nullable, it's either characteristic or error

lib/src/_managers_for_classes.dart
```diff
@@ -101,7 +101,7 @@ abstract class ManagerForService {
     String transactionId,
   );
 
-  Future<Characteristic> writeCharacteristicForService(
+  Future<Characteristic?> writeCharacteristicForService(
```
mikolak: This is not nullable, it's either characteristic or error

lib/src/_managers_for_classes.dart
```diff
@@ -145,7 +145,7 @@ abstract class ManagerForCharacteristic {
     String transactionId,
   );
 
-  Future<void> writeCharacteristicForIdentifier(
+  Future<Characteristic?> writeCharacteristicForIdentifier(
```
mikolak: This is not nullable, it's either characteristic or error

lib/peripheral.dart
```diff
@@ -143,7 +143,7 @@ class Peripheral {
   /// [serviceUuid]. Optional [transactionId] could be used to cancel operation.
   ///
   /// Will result in error if discovery was not done during this connection.
-  Future<Characteristic> writeCharacteristic(
+  Future<Characteristic?> writeCharacteristic(
```
mikolak: This is not nullable, it's either characteristic or error

lib/characteristic.dart
```diff
@@ -69,7 +69,7 @@ class Characteristic extends InternalCharacteristic {
   /// The value can be written only if [isWritableWithResponse] or
   /// [isWritableWithoutResponse] is `true` and argument [withResponse] is
   /// set accordingly.
-  Future<void> write(
+  Future<Characteristic?> write(
```
mikolak: Void here was intentional. There's a bug with not returning response from write with response, but it has to be fixed on multiple layers and even then this would be typed as `Uint8List`.

lib/src/bridge/bluetooth_state_mixin.dart
```diff
@@ -1,9 +1,11 @@
 part of _internal;
 
 mixin BluetoothStateMixin on FlutterBLE {
-  final Stream<dynamic> _adapterStateChanges =
-      const EventChannel(ChannelName.adapterStateChanges)
-          .receiveBroadcastStream();
+  final Stream<String> _adapterStateChanges =
+    const EventChannel(ChannelName.adapterStateChanges)
+      .receiveBroadcastStream()
+      .where((event) => event is String)
```
mikolak: There shouldn't be any filtering on this stream. If for any reason an unexpected value came, it should be raised as an error.

lib/src/bridge/characteristics_mixin.dart
```diff
@@ -11,23 +11,23 @@ mixin CharacteristicsMixin on FlutterBLE {
...
-          .catchError((errorJson) =>
-              Future.error(BleError.fromJson(jsonDecode(errorJson.details))))
+          .catchError((errorJson) async =>
+            throw BleError.fromJson(jsonDecode(errorJson.details))
+          )
           .then((rawValue) {
-            String rawJsonValue = "";
-            if (rawValue is String) {
-              rawJsonValue = rawValue;
+            if (rawValue == null) {
```
mikolak: If characteristic didn't come, then an exception should be raised. Returning some invalid default value only obscures the situation.

lib/src/bridge/characteristics_mixin.dart
```diff
@@ -47,16 +47,16 @@ mixin CharacteristicsMixin on FlutterBLE {
...
-              Future.error(BleError.fromJson(jsonDecode(errorJson.details))))
+          .catchError((errorJson) async =>
+              throw BleError.fromJson(jsonDecode(errorJson.details))
+          )
           .then((rawValue) {
-            String rawJsonValue = "";
-            if (rawValue is String) {
-              rawJsonValue = rawValue;
+            if (rawValue == null) {
+              throw Exception("rawValue cannot be null");
```
mikolak: This exception doesn't give any value to the consumer of the API.

mikolak (changes requested):
I appreciate your work, but please, revert the last commit. Native client has to either return characteristic or error and I'd rather have a null pointer exception thrown than some custom exception that does not give the API consumer any insight or returning an invalid default value.

There's a bug where the response is not properly returned on Android, but it's a multilayer issue that originates in Android's MBA, any changes to interface will be done along with fixing it.

Changes to the mixins are unrelated to this PR and should be in a separate PR.

Summing up: thanks for all the work on null safety! ❤️  Please, run the formatter in scanning_mixin.dart and revert commit `86dafd6` (`Characteristic:: refactor on Futures to always complete...`) and I'll merge it! 🙂

lib/src/bridge/characteristics_mixin.dart
okocsis: That's fine. I understand, you see the correct semantics for this domain much better. It's just that, there were some -- and correct me if I'm wrong where a function returning a future, could become into a hanging state and never actually resolve, only on timeout if the caller sets one. Or is it always guaranteed from the native side that 'some' `.then()` (or `error()`) will always be executed?

The reason I don't get this is that there is this `withResponse` flag user can pass in and in the case of that being false, how will the returned future behave??
To me, it looks like it will never resolve without a(n unnecessary) timeout in the case of `withResponse = false`, and even then it wouldn't return a characteristic in the future raising a _TypeError in new typesafer dart runtime.

and on `withResponse = true` it behaves as expected ofc

This is why I thought it would make sense to make it return a nullable resolving future, that is guaranteed to always resolve, making the caller code much cleaner.

To sum up: 
- I will revert the commit, but pls consider these concerns, as these are changes better be done on a major version bump, as it introduces minor API amendments, potentially breaking adopters code also.
- What does Future<Characteristic> resolve to exactly on `withResponse = false`?

thanks

lib/src/bridge/characteristics_mixin.dart
okocsis: yes, I know. I wanted to discuss what it should throw, but now it's postponed as I've reverted the commit. ;)

lib/src/bridge/characteristics_mixin.dart
mikolak: Response is a byte array from the server (peripheral). If the central (client) does not await for the response, the operation is finished immediately after sending the write request.

Native plugin always returns either characteristic or error. The whole characteristic is returned regardless of the response for convenience - this way you can write to a characteristic from peripheral, save the reference to the characteristic and continue operations on it.

lib/src/bridge/characteristics_mixin.dart
```diff
@@ -20,10 +20,16 @@ mixin CharacteristicsMixin on FlutterBLE {
           )
           .catchError((errorJson) =>
               Future.error(BleError.fromJson(jsonDecode(errorJson.details))))
-          .then((rawJsonValue) =>
-              _parseCharacteristicWithValueWithTransactionIdResponse(
-                      peripheral, rawJsonValue)
-                  .value);
+          .then((rawValue) {
+            String rawJsonValue = "";
+            if (rawValue is String) {
```
mikolak, line 25: This can actually be get rid of by changing to `.invokeMethod<String>(...)`

lib/src/bridge/characteristics_mixin.dart
```diff
@@ -43,11 +49,16 @@ mixin CharacteristicsMixin on FlutterBLE {
...
-          .then(
-            (rawJsonValue) =>
-                _parseCharacteristicWithValueWithTransactionIdResponse(
-                    peripheral, rawJsonValue),
-          );
+          .then((rawValue) {
+            String rawJsonValue = "";
+            if (rawValue is String) {
+              rawJsonValue = rawValue;
+            }
```
mikolak, line 56: This can actually be get rid of by changing to `.invokeMethod<String>(...)`
And yes, it's guaranteed to be string at this point in time.

lib/src/bridge/characteristics_mixin.dart
```diff
@@ -66,11 +77,16 @@ mixin CharacteristicsMixin on FlutterBLE {
...
```
mikolak, line 84: This can actually be get rid of by changing to `.invokeMethod<String>(...)`
And yes, it's guaranteed to be string at this point in time.

lib/src/bridge/characteristics_mixin.dart
```diff
@@ -230,39 +252,47 @@ mixin CharacteristicsMixin on FlutterBLE {
...
-              _parseCharacteristicWithValueWithTransactionIdResponse(
-                  peripheral, rawJsonValue),
-        )
+    Stream<CharacteristicWithValueAndTransactionId> stream =
+      _characteristicsMonitoringEvents
+        .map((rawValue) {
+          String rawJsonValue = "";
+          if (rawValue is String) {
+            rawJsonValue = rawValue;
+          }
```
mikolak, line 267: This can actually be get rid of by changing to `.invokeMethod<String>(...)`
And yes, it's guaranteed to be string at this point in time.

lib/src/internal_ble_manager.dart
```diff
@@ -185,20 +184,20 @@ class InternalBleManager
...
+  Future<int> requestMtu(
       Peripheral peripheral, int mtu, String transactionId) {
     return _bleLib.requestMtu(peripheral, mtu, transactionId);
   }
 
   @override
   Future<List<Peripheral>> knownPeripherals(
       List<String> peripheralIdentifiers) {
-    return _bleLib.knownDevices(peripheralIdentifiers ?? []);
+    return _bleLib.knownDevices(peripheralIdentifiers);
```
mikolak, line 195: Might need confirmation

lib/src/internal_ble_manager.dart
```diff
@@ -185,20 +184,20 @@ class InternalBleManager
...
   }
 
   @override
   Future<List<Peripheral>> connectedPeripherals(List<String> serviceUuids) {
-    return _bleLib.connectedDevices(serviceUuids ?? []);
+    return _bleLib.connectedDevices(serviceUuids);
```
mikolak, line 200: Might need confirmation

mikolak (approved):
A few minors, but I'll take care of those myself. Thank you very much for the PR!

lib/src/bridge/characteristics_mixin.dart
okocsis: yeah, so I figured.  As later I went and explored invokeMethod nullsafe amendments..
I did those on the latest commit that ended up being reverted. :)
we can refactor the whole code with this making it much more type correct! ;)

lib/src/bridge/characteristics_mixin.dart
okocsis: Yes, now I know, I didn't know it then when I wrote this... :/ would have been much more simpler to fixup the bit too loose typing in the code :). But that's how we learn right?

lib/src/internal_ble_manager.dart
okocsis: the confirmation lays underneath ofc not here, but if one can pass in a null then they might as well just pass in an empty list. 
Null on input is not something desirable unless it actually carries a different semantic meaning.
so basically it's just guaranteed by the type system and not this extra safety code.

lib/src/internal_ble_manager.dart
okocsis: same here List<String> serviceUuids doesn't allow a null value

lib/src/bridge/characteristics_mixin.dart
okocsis: > it's guaranteed to be string at this point in time.

As a matter of fact it will have the correct type inferred even of `rawValue` no need to statically denote the type in `then()`-s param list like so:
 - good: `then(rawValue)` inferred to be string cause of invoke's generic type notation
 - bad: `then(String rawValue)` unnecessary and can even break when `invokeMetod<dynamic>`

PR #589, opened 2021-04-09
Add method for checking whether native client exists

### LINKED ISSUE #559: Fetching state of the native client
As it seems, BleManager.createClient should only ever be called once (unless destoryClient is also called) but there is no way to know, if it was already called. BleManager however is a singleton.

Currently I try to solve this with a Wrapper class, that will track this.

Perhaps add a `bool get isCreated` ?

Are there any other patterns how this is usually solved?

Thank you.

mikolak:
Fixes #559 

- [x]  check if it's working properly

### DIFF
--- a/ios/Classes/FlutterBleLibPlugin.m
+++ b/ios/Classes/FlutterBleLibPlugin.m
@@ -71,6 +71,8 @@
         [self createClient:call result:result];
     } else if ([METHOD_NAME_DESTROY_CLIENT isEqualToString:call.method]) {
         [self destroyClient];
+    } else if ([METHOD_NAME_IS_CLIENT_CREATED isEqualToString:call.method]) {
+        [self isClientCreated:call result:result];
     } else if ([METHOD_NAME_ENABLE_RADIO isEqualToString:call.method]) {
         [self enable:call result:result];
     } else if ([METHOD_NAME_DISABLE_RADIO isEqualToString:call.method]) {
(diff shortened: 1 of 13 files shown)


ios/Classes/FlutterBleLibPlugin.m
```diff
@@ -158,7 +160,14 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
 
 // MARK: - MBA Methods - BleClient lifecycle
 
+- (void)isClientCreated:(FlutterMethodCall *)call result:(FlutterResult)result {
+    result([NSNumber numberWithBool:_adapter != nil]);
```
okocsis, line 164: this could just as well be simply `result(@(adapter != nil))`
as there is a literal way to create NSNumber instances 
`@(int|float|boolean)` == [NSNumber numberWith[Int,Float,Bool]:(int|float|boolean).
Similarly to the literal to create NSStrings `@""`

mikolak: Fun fact: they're apparently not the same 😄 
```objc
result(@(_adapter != nil))
```
gives:
> type 'int' is not a subtype of type 'bool?' in type cast

While using `NSNumber....` works. 🤷

mikolak: More info: using literal notation sends 0 to Dart, using NSNumber numberWithBool returns bool.

mikolak (commented):
👍

ios/Classes/FlutterBleLibPlugin.m
okocsis: this is really odd.. https://clang.llvm.org/docs/ObjectiveCLiterals.html

PR #620, opened 2022-02-04
update example

adnanjpg:
I've had a problem with `flutter pub get` with the old version of `rxdart`, updating the version solved it, this may be a common problem so I think it should be up to date.

and I've cleaned some warnings


adnanjpg (commented):
the Example app was refusing to run because of old embedding (V1)

